PR #11, opened 2015-09-17
Fix :Tags command tagfile relative path

atn34:
Since fzf will execute the source command in the tagfile's directory, we
can't give the path of the tagfile relative to the current working
directory. Instead give just the filename.


junegunn:
Thanks!

atn34:
Thanks for the quick response!

PR #19, opened 2015-09-30
Replace the terminal window's statusline in NeoVim

tweekmonster:
Thought this looked nicer in NeoVim and gives it a more distinguished separation.

Before

After

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -76,6 +76,35 @@ function! fzf#complete(...)
   return call('fzf#vim#complete', a:000)
 endfunction
 
+if has('nvim') && get(g:, 'fzf_nvim_statusline', 1)
+  function! s:fzf_restore_colors()
+    if $TERM !~ "256color"
+      highlight fzf1 ctermfg=1 ctermbg=8
+      highlight fzf2 ctermfg=2 ctermbg=8
+      highlight fzf3 ctermfg=7 ctermbg=8
+    else
+      highlight fzf1 ctermfg=161 ctermbg=238
+      highlight fzf2 ctermfg=151 ctermbg=238
+      highlight fzf3 ctermfg=252 ctermbg=238
+    endif
+  endfunction
+
+  function! s:fzf_nvim_term()
+    if get(w:, 'airline_active', 0)
+      let w:airline_disabled = 1
+    endif
+
+    setlocal nospell
+    setlocal statusline=%#fzf1#\ >\ %#fzf2#fz%#fzf3#f
+  endfunction
+
+  augroup fzf_statusline
+    autocmd!
+    autocmd TermOpen *bin/fzf* call s:fzf_nvim_term() | autocmd WinEnter <buffer> call s:fzf_nvim_term()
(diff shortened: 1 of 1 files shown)


junegunn:
Looks pretty, thanks :)

I have some comments:
- Colors are lost when I change color scheme. `highlight` statements probably should be inside the function
- Do we need `BufLeave`? `WinEnter` alone works for me. Am I missing something?
- (I was also wondering if we should consider non-256-color terminals or GUI version of neovim that will appear in the future. I guess it's not really necessary. Text still visible even when TERM is not set to 256-color.)

plugin/fzf.vim
```diff
@@ -76,6 +76,26 @@ function! fzf#complete(...)
...
+  highlight fzf2 ctermfg=157 ctermbg=238
+  highlight fzf3 ctermfg=252 ctermbg=238
+
+  function! s:fzf_nvim_term()
+    if get(w:, 'airline_active', 0)
+      let w:airline_disabled = 1
+    endif
+
+    setlocal nospell
+    exec "setlocal statusline=%#fzf1#%{'\\ \\ \\>\\ '}%#fzf2#%{'fz'}%#fzf3#%{'f'}"
```
junegunn: `exec` is unnecessary; `setlocal statusline=%#fzf1#%{'\ \ \>\ '}%#fzf2#%{'fz'}%#fzf3#%{'f'}` seems to work just fine.

plugin/fzf.vim
```diff
@@ -76,6 +76,26 @@ function! fzf#complete(...)
   return call('fzf#vim#complete', a:000)
 endfunction
 
+if has('nvim') && get(g:, 'fzf_nvim_statusline', 1)
+  highlight fzf1 ctermfg=162 ctermbg=238
+  highlight fzf2 ctermfg=157 ctermbg=238
```
junegunn: The default colors used by fzf are actually 161 and 151.

plugin/fzf.vim
tweekmonster: Oof, thanks!  I was up late and exec was the final form that "just worked"

tweekmonster:
> Colors are lost when I change color scheme. highlight statements probably should be inside the function

I put the highlights in their own function that's ran on VimEnter,ColorScheme

> Do we need BufLeave? WinEnter alone works for me. Am I missing something?

I removed it.  It was a late night and it seemed to cover airline being persistent despite w:airline_disabled being set.

> (I was also wondering if we should consider non-256-color terminals or GUI version of neovim that will appear in the future. I guess it's not really necessary. Text still visible even when TERM is not set to 256-color.)

I put in alternate colors.  I would've used t_Co, but I think NeoVim hardcodes it to 256 regardless of what $TERM is set to.

junegunn:
Looks good to me! Just one more thing: it looks like we can further simplify &statusline as follows

```vim
setlocal statusline=%#fzf1#\ >\ %#fzf2#fz%#fzf3#f
```

Can you update the code and squash the commits into one?

tweekmonster:
Done.  Thanks!

junegunn:
:+1:

PR #20, opened 2015-10-01
Mappings

junegunn:
Not sure about this. Seems almost useless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Commands
 | `Commits`        | Git commits (requires [fugitive.vim][f])                                  |
 | `BCommits`       | Git commits for the current buffer                                        |
 | `Commands`       | Commands                                                                  |
+| `Maps`           | Normal mode mappings                                                      |
 | `Helptags`       | Help tags <sup id="a1">[1](#helptags)</sup>                               |
 
 - Most commands support `CTRL-T` / `CTRL-X` / `CTRL-V` key
@@ -103,6 +104,10 @@ Mappings
 
 | Mapping                            | Description                               |
 | ---                                | ---                                       |
+| `<plug>(fzf-maps-n)`               | Normal mode mappings                      |
+| `<plug>(fzf-maps-i)`               | Insert mode mappings                      |
+| `<plug>(fzf-maps-x)`               | Visual mode mappings                      |
+| `<plug>(fzf-maps-o)`               | Operator-pending mappings                 |
 | `<plug>(fzf-complete-word)`        | `cat /usr/share/dict/words`               |
 | `<plug>(fzf-complete-path)`        | Path completion using `find` (file + dir) |
 | `<plug>(fzf-complete-file)`        | File completion using `find`              |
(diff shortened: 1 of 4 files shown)


netei:
What does it add ?

junegunn:
List of normal mode mappings. But I'm pretty sure I'll never going use it. The only purpose of the command is to showcase (/ brag about) what's possible with fzf.

netei:
I kind of like that idea, to be able to find quickly to which mapping an action is bound, if you tend to forget.

tweekmonster:
I definitely would use this.  I was just thinking about how much I really hate the more-prompt and how unsearchable it is.  Maybe it could be a general function that takes an exec string, arguments for formatting the results, and a callback for dealing with the selection.

Something like: `fzf#vim#execute('let g:')` or a command `FZFExec 'nmap'`

netei:
I like the idea of general function of @tweekmonster

junegunn:
Well actually I'm considering a different approach, `<plug>` map for each mode.

```vim
nmap <leader>? <plug>(fzf-maps-n)
xmap <leader>? <plug>(fzf-maps-x)
omap <leader>? <plug>(fzf-maps-o)
```

Then we can do something like follows:

- `y` - I'm going to yank some ..
- (wait, what was the text object?)
- `<leader>?`
- fzf shows operator-pending maps and you choose one
- yanked!

Or,

- (what was the mapping from vim-commentary?)
- `<leader>?`
- fzf now shows normal mode maps, you select `gc`
- `<leader>?` again
- Pick a text object and the region is commented out

khalidchawtany:
I would love to have this functionality available :+1:

junegunn:
I was thinking it would be much useful if it can show not only custom mappings but also the builtin key bindings. As far as I know, we have to parse relevant sections in `:help index.txt` to do that. (actually it is what `:Commands` currently does)

junegunn:
Merged to master. I'm trying out the following config

```vim
nmap <leader><tab> <plug>(fzf-maps-n)
xmap <leader><tab> <plug>(fzf-maps-x)
omap <leader><tab> <plug>(fzf-maps-o)
```

khalidchawtany:
Is there any way to narrow down the results to the ones that are usable in the current context, like it's in **SpaceMacs**?
For example I have pressed ```di``` and I want to know the mapping for fold text-object. I want to see only stuff which can be used in this context. Some mappings become irrelevant after ```di``` is pressed.

junegunn:
To do that, we'll have to define operator-pending maps for all possible key strokes. I'm not sure if there's a better approach at the moment.

khalidchawtany:
Actually, I have no idea how this can be implemented. However, with context-aware completion this will become super-useful and extra awesome ;)

junegunn:
To be honest, I don't see the point of it. I should know the mapping by heart if I actually use it enough. Not remembering it usually means that I don't use it frequently and I just don't need it. I like to keep it simple and do not like having unused mappings around. My muscle memory knows the mappings that I actually use and I don't need any visual help every time I press the keys. To me it's just annoying and it's the same reason I don't use any fancy, flickering statusline plugins.

khalidchawtany:
Its great as it's. Thanks

junegunn:
Thanks. It's just how I like my editor to be and I wouldn't be surprised if others like it and find it useful. Nothing wrong with it. But anyway, implementing it with the current version of vim will be non-trivial and most likely end up being a huge pile of hack.

PR #34, opened 2015-10-30
Add GitFiles command

icholy:


junegunn:
Thanks. We can already take advantage of `git ls-tree` using `$FZF_DEFAULT_COMMAND` (https://github.com/junegunn/fzf#git-ls-tree-for-fast-traversal), but yeah, I think it's okay to have a dedicated command for it. How about changing the prompt to `GitFiles>`?

junegunn:
And please update doc/fzf-vim.txt file as well.

autoload/fzf/vim.vim
```diff
@@ -324,6 +324,18 @@ function! fzf#vim#history(...)
 endfunction
 
 " ------------------------------------------------------------------
+" GitFiles
+" ------------------------------------------------------------------
+
+function! fzf#vim#gitfiles(...)
+  call s:fzf({
+  \ 'source':  'git ls-tree --name-only --full-tree -r HEAD',
```
junegunn: I don't think we should use `--full-tree`. Vim will unable to find the file if the current working directory is not the root of the repo.

icholy:
Do you think it should it also a take a `[path]` parameter like the `Files` command?

junegunn:
I don't think it's necessary. But I think we should check if we're currently under a git repo and print warning message if not instead of showing the empty result. (like `:Snippets` command when ultisnips is unavailable)

junegunn:
Merged, thanks. I also added the warning message and applied `dir` option which makes fzf start from the root of the repository so that we can see any file on the repository and open it regardless of the current working directory.

PR #36, opened 2015-11-02
Add GitGrep command

icholy:
Is there a way to do this without adding a command?

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Commands
 | ---              | ---                                                                       |
 | `Files [PATH]`   | Files (similar to `:FZF`)                                                 |
 | `GitFiles`       | Git files
+| `GitGrep`       | Git Grep
 | `Buffers`        | Open buffers                                                              |
 | `Colors`         | Color schemes                                                             |
 | `Ag [PATTERN]`   | [ag][ag] search result (`CTRL-A` to select all, `CTRL-D` to deselect all) |
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -376,6 +376,46 @@ function! fzf#vim#buffers(...)
 endfunction
 
 " ------------------------------------------------------------------
+" GitGrep
+" ------------------------------------------------------------------
+function! s:gitgrep_to_qf(line)
+  let parts = split(a:line, ':')
+  return {'filename': parts[0], 'lnum': parts[1], 'col': 0,
+        \ 'text': join(parts[2:], ':')}
+endfunction
+
+function! s:gitgrep_handler(lines)
+  if len(a:lines) < 2
+    return
+  endif
+
+  let cmd = get(get(g:, 'fzf_action', s:default_action), a:lines[0], 'e')
+  let list = map(a:lines[1:], 's:gitgrep_to_qf(v:val)')
+
+  let first = list[0]
+  execute cmd s:escape(first.filename)
+  execute first.lnum
+  execute 'normal!' first.col.'|zz'
+
+  if len(list) > 1
+    call setqflist(list)
+    copen
+    wincmd p
+  endif
+endfunction
+
+function! fzf#vim#gitgrep(query, ...)
+  call s:fzf({
+  \ 'source':  printf('git grep --line-number "%s"',
+  \                   escape(empty(a:query) ? '^(?=.)' : a:query, '"\-')),
+  \ 'sink*':    s:function('s:gitgrep_handler'),
+  \ 'options': '--ansi --delimiter : --nth 4..,.. --prompt "GitGrep> " '.
+  \            '--multi --bind ctrl-a:select-all,ctrl-d:deselect-all '.
+  \            '--color hl:68,hl+:110'.s:expect(),
+  \ }, a:000)
+endfunction
+
+" ------------------------------------------------------------------
 " Ag
 " ------------------------------------------------------------------
 function! s:ag_to_qf(line)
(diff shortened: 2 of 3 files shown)


junegunn:
That looks almost identical to Ag, right? Maybe we could generalize the concept into a versatile command that can be used with `grep`, `ag`, `git grep`, or whatever. I'll think about it. Until then, you can have it on your vimrc, though you'll have to replace `s:` references appropriately.

icholy:
Yeah, I just copied the Ag command. Generalizing it sounds like the right idea.

rainerborene:
What do you think about renaming `Ag` to `Grep` and use `grepprg` and `grepformat` options?

rainerborene:
That would allow users to temporarily change `grepprg` and `grepformat` on a function if they want to.

```VimL
function! s:Ag()
  " store original grepprg and grepformat
  " change grepprg....
  " call grep
  " restore original options
endfunction
```

junegunn:
That's a good point, adding a generalized version of `Ag`. But I'll keep `Ag` just for convenience. You don't have to know how to set up grepprg and grepformat, and it's tad shorter to type :) I'll review the code if you send me a PR.

Incidentally, I have [`Grep` command defined in my vimrc](https://github.com/junegunn/dotfiles/blob/master/vimrc#L1296-L1301) that does not use fzf and is just a shortcut to the builtin `:grep` command. Sometimes in a small repo when you simply want to build a quickfix list of matches, fzf can be an overkill.

LK4D4:
@junegunn Sorry to bother you, but would you be opposed to at least option to replace ag command there? I've tested three tools: `ack`, `ag` and `pt` and all three have same options.
https://gist.github.com/anonymous/8eaa4eff4dc0e248b1d6e9e222088c62 here is the patch which I used for tests.

junegunn:
@LK4D4 That is based on the assumption that the other commands 1. take the same set of options, 2. and print the output in the same format as ag. I'll see if it's viable to add a generalized grep function that takes the full command. The command is also responsible for transforming the output so the same sink function can be applied.

CrashyBang:
Hey @junegunn in regards to the above comment and a generalised grep function will a new issue be opened for this? Asking simply in regards to tracking progress.

Cheers.

junegunn:
@CrashyBang No need. I'm working on it, coming soon.

junegunn:
I'm almost done but there's a few issues.

- Different format. `git grep` or gnu grep do not print the column number unlike ag, pt, or awk
- rg seems unstable at the moment.
    - This does not work on Vim: `echo system('rg ".*"') | echo v:shell_error`

CrashyBang:
Hey @junegunn,

Cheers for letting me know, should I post an issue against `rg`?

Cheers.

icholy:
Lol that's dope

junegunn:
No idea why rg doesn't work well in this context, but it's a separate issue anyway.

This is how the new `fzf#vim#grep(command, with_column, bang_or_extra_opts)` function can be used:

```vim
" git grep and gnu grep do not print column number so with_column is 0
command! -bang -nargs=* GGrep call fzf#vim#grep('git grep --line-number '.shellescape(<q-args>), 0, <bang>0)
command! -bang -nargs=* Grep call fzf#vim#grep('grep -r --line-number '.shellescape(<q-args>).' *', 0, <bang>0)
```

CrashyBang:
Hey @junegunn got this working with `rg` :+1: 

I am using the following command which works perfectly except for one thing:

```vim
command! -bang -nargs=* Find call fzf#vim#grep('rg --vimgrep --fixed-strings --ignore-case --no-ignore --hidden --follow --glob "!.git/*" '.shellescape(<q-args>), 1, <bang>0)
```

It does not highlight the search term like `:Ag` used to, should it highlight it or was that custom to the `:Ag` command?

Cheers,
Otis.

junegunn:
rg seems to ignore `--color=always` if `--vimgrep` is set, which is a sensible thing to do I guess.

So without `--vimgrep`, we can use something like `rg --column --line-number --no-heading --color=always KEYWORD`. It works fine on Neovim (with the latest fzf 0.15.2). But for some reason it doesn't work correctly on Vim running on tmux. I can see some extra `^O` characters in the output, not sure why.

```sh
# Compare these
echo foobar | ag --color foo | more
# ESC[30;43mfooESC[0mESC[Kbar

echo foobar | rg -N --color=always foo | more
# ESC[91mESC[1mfooESC[m^Obar (on tmux)
# ESC[91mESC[1mfooESC(BESC[mbar (not on tmux)
```

CrashyBang:
Hey @junegunn looks good to me, would you like me to open an issue with rg? Or would you be better suited?

junegunn:
Do as you please. FYI, we can make it work without changing rg or fzf.vim by filtering out the character with `tr`.

```vim
command! -bang -nargs=* Find
  \ call fzf#vim#grep('rg --column --line-number --no-heading --color=always '.shellescape(<q-args>).'| tr -d "\017"', 1, <bang>0)
```

rawaludin:
Trying to make it work. But no, this implementation is not faster than @junegunn implementation with Ag in fzf.vim. Just want to make sure, there won't be native :Rg command in this plugin right? 

Thanks.

rawaludin:
Using sample Rg implementation in Readme, when I'm using :Rg its count total files longer than using :Ag command. The result is come late too.

rawaludin:
See here https://asciinema.org/a/4ab9hb2mxkmlc7ig02naufl0h

junegunn:
I don't know, I just use `:Ag KEYWORD` :) The performance of ag has never been a problem for me, so I haven't really looked into rg.

rawaludin:
Yeah, better stick with Ag :)

stephenprater:
@rawaludin I'm having this same problem - did you ever figure it out?  Running `rg` is twice as fast as `ag` just running it from bash - but running within FZF.vim and `rg` is _way_ slower than `ag`.

Issue #33, opened 2015-10-30
De-duplicate :History


tlvince:
`:History` contains many duplicated files, for example `package.json`:

![screen shot 2015-10-30 at 11 20 26](https://cloud.githubusercontent.com/assets/323761/10844560/5d6ad5b8-7ef8-11e5-9afe-af0a0b4b91cd.png)

Can these be de-duplicated?

junegunn:
I don't have the problem. Can you post the output of `:oldfiles` and `:buffers` when that happens?

tlvince:
```
:oldfiles
1: /usr/local/Cellar/neovim/HEAD/share/nvim/runtime/doc/windows.txt
2: bower.json
3: gulp/inject.js
4: .bowerrc
5: gulp/inject.js
6: node_modules/couchdb-bootstrap/lib/configure.js
7: /usr/local/Cellar/neovim/HEAD/share/nvim/runtime/doc/eval.txt
8: /usr/local/Cellar/neovim/HEAD/share/nvim/runtime/doc/starting.txt
9: /usr/local/Cellar/neovim/HEAD/share/nvim/runtime/doc/eval.txt
10: node_modules/couchdb-bootstrap/lib/secure.js
11: node_modules/couchdb-bootstrap/lib/configure.js
12: node_modules/couchdb-bootstrap/node_modules/couchdb-secure/index.js
13: node_modules/couchdb-compile/package.json
14: node_modules/couchdb-compile/index.js
15: node_modules/couchdb-compile/package.json
16: node_modules/couchdb-bootstrap/node_modules/couchdb-configure/index.js
17: node_modules/couchdb-bootstrap/lib/configure.js
18: node_modules/couchdb-bootstrap/index.js
19: node_modules/couchdb-bootstrap/node_modules/nano-option/index.js
20: node_modules/couchdb-bootstrap/node_modules/nano-option/package.json
21: node_modules/couchdb-bootstrap/node_modules/nano-option/index.js
22: node_modules/couchdb-bootstrap/index.js
23: node_modules/couchdb-ensure/index.js
24: node_modules/couchdb-ensure/node_modules/nano-option/index.js
25: gulp/couchdb.js
26: node_modules/couchdb-ensure/node_modules/nano-option/node_modules/nano/lib/nano.js
27: node_modules/couchdb-ensure/node_modules/nano-option/index.js
28: node_modules/couchdb-ensure/index.js
29: gulp/couchdb.js
30: gulp/couchdb.js
[...]

PR #57, opened 2015-12-25
fzf#vim#history: only use v:oldfiles

blueyed:
This fixes having duplicates in the list.

Fixes https://github.com/junegunn/fzf.vim/issues/33.


blueyed:
Closes, see https://github.com/junegunn/fzf.vim/issues/33#issuecomment-167243581.

Issue #51, opened 2015-12-05
Supply additional arguments to :Ag


danihodovic:
Is it possible to supply additional arguments to `:Ag` such as `-Q` and `-t` in order to search .gitignored files?  I'm guessing that the `source` dict key decides the source of `:Ag` and that can't be modified.
https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L404-L405

junegunn:
Sounds like a reasonable suggestion. Currently `:Ag -Q foobar` will search for `-Q foobar` as the query. If we make `:Ag` command take command-line options, it will be impossible to support queries that start with a dash without escaping of it. e.g.`:Ag --foobar`

junegunn:
```patch
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index a38a4b9..2ccb63e 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -400,9 +400,19 @@ function! s:ag_handler(lines)
 endfunction
 
 function! fzf#vim#ag(query, ...)
+  let ag_opts = ''
+  let query = ''
+  for token in split(a:query, '\s\+\zs')
+    if empty(query) && token =~ '^-'
+      let ag_opts .= token
+    else
+      let query .= token
+    endif
+  endfor
+
   return s:fzf(fzf#vim#wrap({
-  \ 'source':  printf('ag --nogroup --column --color "%s"',
-  \                   escape(empty(a:query) ? '^(?=.)' : a:query, '"\-')),
+  \ 'source':  printf('ag --nogroup --column --color %s "%s"',
+  \                   ag_opts, empty(query) ? '^(?=.)' : query),
   \ 'sink*':    s:function('s:ag_handler'),
   \ 'options': '--ansi --delimiter : --nth 4..,.. --prompt "Ag> " '.
   \            '--multi --bind ctrl-a:select-all,ctrl-d:deselect-all '.
```

This is something we can try. With this, one will have to escape queries starting with a dash, e.g. `:Ag \--foobar`, or it will be passed as an ag option. One issue I noticed with the change is you can't use `:Ag -Q`, since in that case the dummy regular expression we use to match every line no longer works.

danihodovic:
[...]

PR #62, opened 2016-01-05
Pull out the git log command as an option

telemenar:
In my environment '--graph' is a kiss of death. It results in seeing
nothing but graph lines due to the number of branches in the environment.

Secondly the colors baked into this break with my color scheme
which has a background of black rending the date unreadable.

I pulled it out as an option to just specify the command. This might be too flexible but it was an easy solution for me.  I just add to my ``.vimrc``:

```
let g:fzf_commit_log_command = 'git log --color=always --format="%C(auto)%h%d %s %C(blue)%C(bold)%cr"'
```

### DIFF
--- a/doc/fzf-vim.txt
+++ b/doc/fzf-vim.txt
@@ -115,7 +115,7 @@ But its functionality is still available via `call pathogen#helptags()`.)
 < Customization >_____________________________________________________________~
                                                          *fzf-vim-customization*
 
-                                                     *g:fzf_action* *g:fzf_layout*
+                           *g:fzf_action* *g:fzf_layout* *g:fzf_commits_log_options*
 >
     " This is the default extra key bindings
     let g:fzf_action = {
@@ -125,6 +125,10 @@ But its functionality is still available via `call pathogen#helptags()`.)
 
     " Default fzf layout
     let g:fzf_layout = { 'down': '40%' }
+
+    " For Commits and BCommits to customize the options used by 'git log':
+    let g:fzf_commits_log_options = \
+        '--graph --color=always --format="%C(auto)%h%d %s %C(black)%C(bold)%cr"'
 
     " Advanced customization using autoload functions
     autocmd VimEnter * command! Colors
--- a/README.md
+++ b/README.md
@@ -97,6 +97,9 @@ let g:fzf_action = {
 " - window (nvim only)
 let g:fzf_layout = { 'down': '~40%' }
 
+" For Commits and BCommits to customize the options used by 'git log':
+let g:fzf_commits_log_options = '--graph --color=always --format="%C(auto)%h%d %s %C(black)%C(bold)%cr"'
+
 " Advanced customization using autoload functions
 autocmd VimEnter * command! Colors
   \ call fzf#vim#colors({'left': '15%', 'options': '--reverse --margin 30%,0'})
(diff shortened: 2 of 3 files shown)


simonweil:
+1

junegunn:
Thanks. Adding configuration variables is the last thing I want to do, but unfortunately I don't see a better option here. (One thing leads to another, there were similar requests regarding `:Ag` command, we might also look into it after this. #51)

- Considering that this repository is a collection of a bunch of unrelated commands, we should have a clear, straightforward, and consistent naming scheme for variables. I suggest that we prefix each variable with the name of the command it is for, so in this case `g:fzf_commits_something`

- As you mentioned, the option can be too flexible. Users might get an impression that they can use any command, such as custom wrapper scripts in place of `git log`, but it is not true as `:BCommits` will append `--follow` option which is only guaranteed to work with `git log`. So how about we limit the scope of customizability as follows:
```vim
let g:fzf_commits_log_options = '--color=always --format="%C(auto)%h%d %s %C(blue)%C(bold)%cr"'
```
If you have better ideas, please let me know.

telemenar:
Added a commit to switch over to specifying options. 

I might open another one for ``:Tags``/``:BTags`` as well. I really like to be able to specify ``--extra=+q`` for that. Unlike ``:Commits`` though with ``:Tags`` I can generate my own file separately and ``:BTags`` is still usable without the extra options I prefer.

I think the Ag issue in #51 is slightly more complicated as what seems like the most obvious solution is either developing a shadow hierarchy of options similar to the ones that interact with the system ``:grep`` and then also interactions/similarities to installs of https://github.com/rking/ag.vim or https://github.com/mileszs/ack.vim

junegunn:
Thanks, can you squash the commits into one?

telemenar:
Done.

doc/fzf-vim.txt
```diff
@@ -126,10 +126,13 @@ But its functionality is still available via `call pathogen#helptags()`.)
...
     let g:fzf_layout = { 'down': '40%' }
 
+    " For Commits and BCommits to customize the options used by 'git log':
+    let g:fzf_commits_log_options = \
+        '--graph --color=always --format="%C(auto)%h%d %s %C(black)%C(bold)%cr"'
+
     " Advanced customization using autoload functions
     autocmd VimEnter * command! Colors
       \ call fzf#vim#colors({'left': '15%', 'options': '--reverse --margin 30%,0'})
-<
```
junegunn: I don't think this should go, right?

telemenar: Good catch, updated again with fix.

doc/fzf-vim.txt
```diff
@@ -115,7 +115,7 @@ But its functionality is still available via `call pathogen#helptags()`.)
 < Customization >_____________________________________________________________~
                                                          *fzf-vim-customization*
 
-                                                     *g:fzf_action* *g:fzf_layout*
+                            *g:fzf_action* *g:fzf_layout* *g:fzf_commit_log_command*
```
junegunn: Ah sorry for not pointing it out earlier, but this also should be updated.

telemenar:
helptag name updated.

junegunn:
Merged, thanks.

PR #87, opened 2016-02-19
fzf#vim#buffer_tags([tag commands], options)

AdnoC:
Added ability to specify the commands used to generate tags
Some filetypes have better tag generators than ctags. For example, Javascript has jsctags which found all the functions in a file while ctags found 4.

This change lets you set the BTags command to use whatever commands you want to generate tags.
```
autocmd FileType javascript command! -buffer BTags
  \ call fzf#vim#buffer_tags([
    \ printf('jsctags %s -f', expand('%:S')),], {})
```

These changes are based on https://github.com/junegunn/fzf.vim/commit/0ad88dd4e8136e7eed48c27937a1f55efe0bc69c

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -432,14 +432,12 @@ endfunction
 " ------------------------------------------------------------------
 " BTags
 " ------------------------------------------------------------------
-function! s:btags_source()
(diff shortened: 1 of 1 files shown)


netei:
+1, think it could be useful

junegunn:
Merged, thanks.

PR #88, opened 2016-02-20
Add :Sessions command

awinecki:
Junegunn, thank you for your awesome work. FZF basically reinvented my flow both in shell and vim.

I'm suggesting adding **:Sessions** command for handling vim native sessions.

**Background**

My goal was to kind of reproduce fast project switching that's available in editors like Atom, Sublime Text (and many IDEs). I've been trying to setup a nice flow working with vim native sessions. The default behaviour is not very handy, so I tried to fix it. Finally, my setup is this:

- https://github.com/mhinz/vim-startify: easy access to most recently opened sessions, also provides couple handy commands, like **:SSave**, **:SLoad** for managing them, and keeps session files to be saved in `~/.vim/session` directory,
- https://github.com/tpope/vim-obsession: provides **:Obsess** command, which handles auto-saving sessions so you basically never worry about them after they're created,
- `alias vs='vim -S "~/.vim/session/$(ls ~/.vim/session | fzf)"'` for quick launching vim with fuzzy selected session,
- `:Sessions` command mapped to handy keyboard shortcut for quick switching between session from inside vim.

I tried to go through the documentation and implement this using the fzf completion helper you'd provided, but without success. Finally, I've added `:Sessions` command, which seemed a bit cleaner. Anyway, there might be a simpler way to do this which I don't see (I'd welcome any hints).

Now, the code I've provided is obviously a no-go. The `source` parameter should not have the sessions dir hardcoded, it should be a global optional variable, something like `let g:fzf_sessions_directory`. Secondly, the `SLoad` command is specific to Startify plugin. Since vim sessions can simply be sourced, there's probably a way to do just that, avoiding any external dependencies.

**Disclaimer**: I've not written any vimscript before, so there's probably better ways to write this code. If you indeed decide that such command could be added, I'd be happy to modify my code to make that be mergeable, if you provide me some hints :).

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -269,6 +269,17 @@ function! fzf#vim#locate(query, ...)
 endfunction
 
 " ------------------------------------------------------------------
+" Sessions
+" ------------------------------------------------------------------
+function! fzf#vim#sessions(...)
+  call s:fzf({
+  \ 'source':  'ls -1 ~/.vim/session',
+  \ 'sink':  'SLoad',
+  \ 'options': '+m --prompt="Sessions> "'
+  \}, a:000)
+endfunction
+
+" ------------------------------------------------------------------
 " History[:/]
 " ------------------------------------------------------------------
 function! s:all_files()
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -60,6 +60,7 @@ call s:defs([
 \'command! -bang Commits                       call fzf#vim#commits(s:w(<bang>0))',
 \'command! -bang BCommits                      call fzf#vim#buffer_commits(s:w(<bang>0))',
 \'command! -bang Maps                          call fzf#vim#maps("n", s:w(<bang>0))',
+\'command! -bang Sessions                      call fzf#vim#sessions(s:w(<bang>0))',
 \'command! -bang -nargs=* History              call s:history(<q-args>, <bang>0)'])
 
 function! s:history(arg, bang)


junegunn:
I'm not familiar session support of vim-startify. I'll take a look and see what we can do.

junegunn:
Hi, I've taken a look and it looks like it's not easy to remove startify requirement. Vim allows you to save and load sessions as files but the concept of sessions directory is not native and is tied to startify. So I decided not to merge this, but you can still have it on your vimrc like follows:

```vim
function! s:sessions(...)
  call fzf#run({
  \ 'source':  'ls -1 ~/.vim/session',
  \ 'sink':    'SLoad',
  \ 'options': '+m --prompt="Sessions> "',
  \ 'down':    '40%'
  \})
endfunction
command! Sessions call s:sessions()
```

Thanks for your suggestion.

awinecki:
Yeah, I agree that's the right way to go here. Any advanced session management seems out of fzf.vim scope.

It works perfectly sitting in my .vimrc as you suggested.

Maybe we could add some more examples in the docs of such unorthodox uses of the plugin? I feel it's really powerful what you can do with the plugin, basically "fuzzify" your whole VIM workflow, but it's not a straightforward thing to do at first.

junegunn:
> such unorthodox uses of the plugin?

It's hard to tell it's unorthodox. Since before this project was created, we have had these pages:

- https://github.com/junegunn/fzf#usage-as-vim-plugin
- https://github.com/junegunn/fzf/wiki/Examples-(vim)

And the users had to learn how to write their own functions and commands using `fzf#run()` function. Now with this fzf.vim project, most "non-hardcore" users don't have to go through the step, but that doesn't mean this project is trying to completely replace direct uses of `fzf#run()` function. I only see it as a maintained version of the aforementioned wiki page.

It seems that some new users who started using fzf.vim from the beginning do not know about the API. We can provide links to the pages.

Issue #109, opened 2016-03-31
Searching through file types.


majjoha:
The clean buffer normally has no file type set. It would be quite convenient if `setf`, and FZF could be coupled, so that one could search through the possible file types to select for the new buffer. Is this somehow possible?

PR #110, opened 2016-04-03
Add `Filetypes` command.

majjoha:
This pull request introduces a new command that allows you to search through the various valid file types.

It essentially fixes the issue I created here: https://github.com/junegunn/fzf.vim/issues/109.


junegunn:
Thanks. Why don't we look at `syntax/*.vim` instead?

majjoha:
Good idea! I've updated the pull request.

junegunn:
Thanks.

PR #111, opened 2016-04-03
add neovim terminal files to history filter

low-ghost:
terribly small change, but pulls all neovim terminal files from the history file search via pattern `term://`


junegunn:
Thanks.

PR #115, opened 2016-04-11
Add section detailing how to install ff.vim with other plugin managers

faceleg:
See https://github.com/Shougo/dein.vim/issues/74#issuecomment-208141136

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Plug 'junegunn/fzf.vim'
 
 - Use `./install --bin` instead if you don't need fzf outside of Vim
 - Make sure to use Vim 7.4 or above
+- See (installation tips)[#installation-tips] for information on installing with different plugin managers
 
 Commands
 --------
@@ -221,6 +222,16 @@ function! s:fzf_statusline()
 endfunction
 
 autocmd! User FzfStatusLine call <SID>fzf_statusline()
+```
+
+Installation Tips
+-----------------
+
+### dein.vim
+
+```
+call dein#add('junegunn/fzf', { 'build': './install', 'merged': 0 })
+call dein#add('junegunn/fzf.vim')
 ```
 
 License


README.md
```diff
@@ -223,6 +224,16 @@ endfunction
 autocmd! User FzfStatusLine call <SID>fzf_statusline()
 ```
 
+Installation Tips
+-----------------
+
+### dein.vim
+
+```
+call dein#add('junegunn/fzf', { 'build': './install', 'merged': 0 })
```
wsdjeg: this only works for unix os,maybe windows has different stuff

faceleg: Agree, hopefully if this gets merged a kind windows user will come along and update this section

kdelchev: README has this line "Also note that fzf currently does not compile on Windows."

faceleg:
@junegunn bump?

junegunn:
fzf and fzf.vim follow standard pathogen-compatible directory layout and should be installable without any special instructions. For example for Vundle, they can be specified just like any other ordinary Vim plugins:

```vim
Plugin 'junegunn/fzf'
Plugin 'junegunn/fzf.vim'
```

For vim-plug it's just:

```vim
Plug 'junegunn/fzf'
Plug 'junegunn/fzf.vim'
```

The README page suggests `dir` and `do` options but they are in fact optional as the plugin automatically downloads the binary if it's not found on the system.

I believe the same goes for NeoBundle, VAM, etc. They're all similar. To be honest, I do not know well enough about dein and its `merged` option you're suggesting here. If dein requires some special configuration for a plugin that other plugin managers don't have any issue with, it is a problem dein has to address.

faceleg:
Thanks!

junegunn:
No problem, thank you for your interest in the project.

PR #117, opened 2016-04-12
Make sure to use the cterm color attribute

buztard:
This should fix color selection with true-color enabled terminal neovim.


junegunn:
👍

PR #120, opened 2016-04-13
Support for switching airline themes

ExarchD:
Hi,
I thought this might be useful, it basically just copies the colorscheme function with minor changes, to allow aiirline theme swithcing. However, not everyone uses airline, so I can appreciate that you might not want to introduce bloat.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -281,6 +281,17 @@ function! fzf#vim#colors(...)
 endfunction
 
 " ------------------------------------------------------------------
+" AirlineThemes
+" ------------------------------------------------------------------
+function! fzf#vim#AirlineTheme(...)
+  return s:fzf({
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but like you said, not everyone uses airline (I don't) and I'd like to avoid adding commands that depend on other plugins.

I suppose you already know but you can just put it in your Vim configuration:

```vim
command! AirlineThemes call fzf#run({
  \ 'source':  map(split(globpath(&rtp, 'autoload/airline/themes/*.vim'), "\n"),
  \               "substitute(fnamemodify(v:val, ':t'), '\\..\\{-}$', '', '')"),
  \ 'sink':    'AirlineTheme',
  \ 'options': '+m --prompt="Airline Themes> "',
  \ 'down':    '~40%'
  \})
```

roma-guru:
Just one more thing - random theme, useful for discoveries:
```vim
\ 'source': ["random"] + map(split(globpath(&rtp, 'autoload/airline/themes/*.vim'), "\n"),
```

PR #127, opened 2016-04-18
Setting to prevent automatic tag generation.

robhurring:
I'm currently using [vim-gutentags](https://github.com/ludovicchabant/vim-gutentags) to manage my tagfiles and would like to prevent fzf.vim from trying to create them for me.

I've added the following setting that can disable that behavior:

    let g:fzf_generate_tags = 0


junegunn:
Thanks. I have felt that immediately generating tags file is not a good default behavior especially when you're not at the root of a project. Instead of adding an option, I'll make the command prompt the user for confirmation.

PR #131, opened 2016-04-20
Add GitChanged command

wezm:
Lists staged and unchanged files tracked by git.


junegunn:
Hi, thanks for the input. It's actually something I've been thinking for a while recently. I'd like to use `git status` instead as it shows the status of each entry which is nice, and I'd rather implement it as an alternative behavior of `:GitFiles` command instead of adding a new command. I'll push a commit shortly.

junegunn:
Added `:GitFiles?` in 9e2aa1a710d0e04fc7707b8879b1df56aaf22fd3. Thanks.

wezm:
Excellent, that's even better. Thanks.

PR #132, opened 2016-04-21
removed unneeded slice

andviro:
Hi there! 
There's a small bug in FZFLines that prevents jumping to line found in another buffer. I've removed the slicing and now it works ok.


junegunn:
Thanks!

PR #136, opened 2016-04-26
lines: display buffer names if there are more than 100 columns

blueyed:
This also uses --tabstop=1 explicitly, where the default of 8 would add
too much whitespace - especially with another tabstop being used to
split the line number from the buffer name.

What do you think?


junegunn:
Tried it and it's really cool. Thanks!

PR #137, opened 2016-04-26
Add simple caching in fzf#vim#_lines

blueyed:
This improves the speed when calling it a second time (and buffers have
not changed).  Not sure if it is worth the potential memory consumption.

If this is feasible, the cache could be invalidated through some
autocommand maybe?!


blueyed:
I think it's not really worth it, and would have to keep e.g. `b:changedtick` into account, too.

junegunn:
Right, let's not enter the world of cache invalidation.

PR #140, opened 2016-04-28
[vim] handle 'nowrapscan' option in s:tags_sink

blueyed:
With 'nowrapscan' a tag might not be found, depending of the position in
the file.


blueyed:
This should probably also get wrapped in a `try … finally` block, but I've seen some weird behavior in that case.  Will give it a shot again now.

junegunn:
Thanks!

Issue #27, opened 2015-10-16
Is there a way to pass a path to :Ag?


craigp:
This is about the only thing I miss from ag.vim, I used it often to search only sub-paths in my project. I looked at the fzf plugin source, doesn't look like it's there at the moment, would it be very difficult to add this?

One thing I like about this version is that I no longer have to include flags (like `:Ag -Q "blah" web/server`) but as I see it that might make it difficult to pass it a path. Problem is that the search/reduce once the :Ag pane is open doesn't also search against the filename/path 

EDIT: just an edit to say thanks so much for fzf and this plugin, which removed a lot of stuff from my vimrc :)

junegunn:
> One thing I like about this version is that I no longer have to include flags

Yeah, that and we can easily pass pattern with spaces `:Ag foo bar`.

> doesn't also search against the filename/path

I also thought about lifting the restriction. Do you still think you're going to need path argument without the restriction?

craigp:
> I also thought about lifting the restriction. Do you still think you're going to need path argument without the restriction?

Probably not actually, that would be enough

junegunn:
So I changed the `nth` option for `Ag` command to `4..,..` from `4..`, which means fzf first matches without the file/line/col part, but looks at the the whole string if no match is found. It's a way to give more priority to the matches in the file content. The trick has its limitations but let's see how it goes.

craigp:
Thanks!

rainerborene:
This also works for me. Only problem is `:Ag` cannot search hidden files.

blueyed:
It would still be very useful to search in a different path, without having to `:(l)cd` before; while it is useful and worked around for subdirs, the workaround does not work for other dirs (that are not below the current path).
Maybe a separate command could be used, e.g. `:AgIn` or something similar?

rainerborene:
@junegunn what do think about allowing the `ag` command to be changed instead?

```diff
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index 445faa0..eb6a85e 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -403,10 +403,10 @@ endfunction
 function! fzf#vim#ag(query, ...)
   let args = copy(a:000)
   let ag_opts = len(args) > 1 ? remove(args, 0) : ''
+  let cmd = get(g:, 'fzf_ag_command', 'ag --nogroup --column --color %s "%s"')
+
[...]

PR #144, opened 2016-05-15
Add g:fzf_ag_raw option

peanutandchestnut:
Don't know how to set search path with Ag, it's also not convenient to pass extra option. So i make a little change.  Default Ag behavior is unchanged  except now -complete=dir
You can toggle on ag raw input by
```Vim
let g:fzf_ag_raw = 1
```
Although old codes will break if you use function to pass extra option to Ag. But i think people who pass extra options should be used to ag default behavior. 

By the way, i like your job. With fzf, i don't need ctrlp, tagbar, Ag.vim anymore, and everything is so fast.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,10 @@ let g:fzf_commits_log_options = '--graph --color=always --format="%C(auto)%h%d %
 
 " [Tags] Command to generate tags file
 let g:fzf_tags_command = 'ctags -R'
+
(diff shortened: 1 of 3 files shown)


junegunn:
Have you seen the discussion in #27? You may find [`AgIn` variation](https://github.com/junegunn/fzf.vim/issues/27#issuecomment-185761539) interesting.

PR #146, opened 2016-05-24
Add GitFiles/ command

JPricey:
Lists all files not ignored by git, including untracked files


junegunn:
Thanks, but `GitFiles/` you added is identical to `GitFiles` except for its choice of `ls-files` options. What if someone wants a slightly different set of options? Therefore, I'd rather add a configuration variable or make `GitFiles` take the options as its arguments; e.g. `:GitFiles -co --exclude-standard`. One can define a mapping or a shortcut command.

PR #150, opened 2016-06-05
[GitFiles] Use system instead of systemlist

dguo:
`systemlist` doesn't exist in older versions of Vim. See jebaum/vim-tmuxify#18

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -405,7 +405,7 @@ function! s:git_status_sink(lines) abort
 endfunction
 
 function! fzf#vim#gitfiles(args, ...)
-  let root = systemlist('git rev-parse --show-toplevel')[0]
+  let root = split(system('git rev-parse --show-toplevel'), '\n')[0]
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -405,7 +405,7 @@ function! s:git_status_sink(lines) abort
 endfunction
 
 function! fzf#vim#gitfiles(args, ...)
-  let root = systemlist('git rev-parse --show-toplevel')[0]
+  let root = split(system('git rev-parse --show-toplevel'), '(\n|\r\n)')[0]
```
junegunn: Can you change `'(\n|\r\n)'` to simply `'\n'`? The regex of vim is eccentric and requires you to escape parentheses, and considering that fzf does not run on Windows, I think `'\n'` should suffice.

dguo:
No problem. I made the change. And that's good to know! I've never touched vimscript before.

junegunn:
Merged, thanks!

Issue #92, opened 2016-02-25
passing options to `:Ag`


ninrod:
skimming through the source, I was positively sure that I could do `:Ag foo --hidden` and the `--hidden` option would be passed down to the ag command. But it is not working. 

Maybe this is a bug?

junegunn:
No, it's how it's supposed to work currently. You'll have to define your own command using `fzf#vim#ag(...)` function. See #51 for the previous discussion on the topic. I'll probably provide a sample implementation as the question seems to crop up frequently.

ninrod:
Ok. Perhaps I'm missing something. I understand that [this snippet][1] will grab `--hidden` if I ex command call `:Ag foo --hidden` and put it inside variable `ag_opts`. Then `ag_opts` gets passed down to the `printf` construct on line `424` which should produce an `ag` call with the `--hidden` option. What am I overlooking here?

[1]: <https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L419-L425>

junegunn:
The actual command that calls the function:
https://github.com/junegunn/fzf.vim/blob/master/plugin/fzf.vim#L52

ninrod:
I see. I understand that I'm suppose to rewrite the [fzf#vim#ag][1] function in my vimrc. Is that correct?

[1]: <https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L419-L430>

junegunn:
Not exactly, you can do that, but it'll be much easier to just use it in your command.

```vim
function! s:ag_with_opts(arg, bang)
  let tokens  = split(a:arg)
  let ag_opts = join(filter(copy(tokens), 'v:val =~ "^-"'))
  let query   = join(filter(copy(tokens), 'v:val !~ "^-"'))
  call fzf#vim#ag(query, ag_opts, a:bang ? {} : {'down': '40%'})
endfunction

autocmd VimEnter * command! -nargs=* -bang Ag call s:ag_with_opts(<q-args>, <bang>0)
```

caneta:
I don't know VimL, but I've modified the `ag_with_opts` function you provided above, adding a default argument `--java` and a different dir:
```
function! s:ag_with_opts(arg, bang)
  let tokens  = split(a:arg)
  call insert(tokens, '--java')
  let ag_opts = join(filter(copy(tokens), 'v:val =~ "^-"'))
[...]

PR #164, opened 2016-06-17
add files completion to Ag

ryanong:


junegunn:
Can you elaborate? Ag command takes a search query as its argument not a name of a file.

junegunn:
Is it because you find it useful to have file names easily accessible on command line even though it's not semantically correct?

ryanong:
Ag accepts paths beyond the first argument to search. So if I want to search a specific directory or file I use that as the second argument

junegunn:
No, `Ag` from this repository does not accept any path argument. The whole string, including white spaces, is passed to `ag` command as the query. So I'm concerned that enabling file completion might confuse the users. It has been suggested that one who wants to customize the behavior should define a custom command using the underlying autoload function. See #27, #92, #144 for the details.

junegunn:
Please consider defining your own command using `fzf#vim#grep` function.

```vim
command! -bang -nargs=* -complete=dir Dag
  \ call fzf#vim#grep('ag --nogroup --column --color '.join(map([<f-args>], 'shellescape(v:val)')), 0, <bang>0)
```

PR #190, opened 2016-08-31
Ignore generated doc/tags

aureooms:


aureooms:
see https://github.com/junegunn/fzf.vim/pull/191

PR #197, opened 2016-09-11
Open selected target with ctrl-o

akupila:
I moved from ctrl-p to fzf and liked this behavior there. You browse results with `ctrl-j/k` and open the current one with `ctrl-o`


junegunn:
Thanks, but since we can already customize the keys via `g:fzf_action`, I would rather not change the default. See https://github.com/junegunn/fzf.vim#global-options

PR #198, opened 2016-09-15
Added AgHistory command

AshyIsMe:
I've added the AgHistory command which is similar to the Ag command except that it searches through the contents of `v:old_files` (not just the filenames in `v:old_files`).

This essentially adds a "grep recent files" use case.

Please critique my implementation, Is there a better way to do with with the existing `fzf#vim#ag()` function?

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,30 +47,31 @@ Plug 'junegunn/fzf.vim'
 Commands
 --------
 
-| Command          | List                                                                      |
-| ---              | ---                                                                       |
-| `Files [PATH]`   | Files (similar to `:FZF`)                                                 |
-| `GitFiles`       | Git files                                                                 |
-| `Buffers`        | Open buffers                                                              |
-| `Colors`         | Color schemes                                                             |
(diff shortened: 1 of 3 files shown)


junegunn:
Thanks, it's a nice idea but I'm currently not interested in adding more commands to this repository as the main purpose of it is simply to give the idea what can be built using fzf on Vim rather than to provide different variations of similar features for different requirements.

Anyway in this case, I believe you can use `fzf#vim#ag_raw` function as follows.

```vim
command! -nargs=+ AgHistory call fzf#vim#ag_raw(shellescape(<q-args>).' '.
\ join(map(filter(copy(v:oldfiles), 'filereadable(expand(v:val))'), 'shellescape(expand(v:val))')),
\ {'down': '40%'})
```

AshyIsMe:
Ah yeap, no worries.

For anyone else who might want something similar this is what I ended up using:
```viml
command! AgHistory call fzf#vim#ag_raw(shellescape('^(?=.)').' '.
\ join(map(filter(copy(v:oldfiles), 'filereadable(expand(v:val))'), 'shellescape(expand(v:val))')),
\ {'down': '40%'})

```

It's almost the same as @junegunn's suggestion but requires no arguments.

PR #202, opened 2016-09-22
s:tags_sink(): Use \M instead of 'nomagic'.

justinmk:
Also removed setting of 'wrapscan', I wouldn't think it is needed in this context.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -627,15 +627,15 @@ function! s:tags_sink(lines)
   normal! m'
   let qfl = []
   let cmd = get(get(g:, 'fzf_action', s:default_action), a:lines[0], 'e')
-  let [magic, &magic, wrapscan, &wrapscan] = [&magic, 0, &wrapscan, 1]
+  let [wrapscan, &wrapscan] = [&wrapscan, 1]
   for line in a:lines[1:]
     let parts = split(line, '\t\zs')
-    let excmd = matchstr(join(parts[2:], ''), '^.*\ze;"\t')
+    let excmd = matchstr(join(parts[2:], ''), '\M^.*\ze;"\t')
     call s:open(cmd, parts[1][:-2])
     execute excmd
     call add(qfl, {'filename': expand('%'), 'lnum': line('.'), 'text': getline('.')})
   endfor
-  let [&magic, &wrapscan] = [magic, wrapscan]
+  let &wrapscan = wrapscan
   if len(qfl) > 1
     call setqflist(qfl)
     copen
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -627,7 +627,6 @@ function! s:tags_sink(lines)
   normal! m'
   let qfl = []
   let cmd = get(get(g:, 'fzf_action', s:default_action), a:lines[0], 'e')
-  let [wrapscan, &wrapscan] = [&wrapscan, 1]
   for line in a:lines[1:]
     let parts = split(line, '\t\zs')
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -627,15 +627,13 @@ function! s:tags_sink(lines)
   normal! m'
   let qfl = []
   let cmd = get(get(g:, 'fzf_action', s:default_action), a:lines[0], 'e')
-  let [magic, &magic, wrapscan, &wrapscan] = [&magic, 0, &wrapscan, 1]
   for line in a:lines[1:]
     let parts = split(line, '\t\zs')
-    let excmd = matchstr(join(parts[2:], ''), '^.*\ze;"\t')
+    let excmd = matchstr(join(parts[2:], ''), '\M^.*\ze;"\t')
```
justinmk: With `\M` (or `'nomagic'`) the `*` here is literal. Is that intentional?

junegunn:
> Before this change if the pattern is not found, the user's 'magic' will not be restored.

Okay, I can see that happens when the tags file is not in sync with the real file.

> We aren't searching a buffer

We are. `excmd` holds an EX command in the form of `/some-pattern-to-locate-the-position-in-the-file/`. `&wrapscan` was suggested by @blueyed in #140.

> With \M (or 'nomagic') the * here is literal. Is that intentional?

`set nomagic | echo 'abc' =~ 'a*'` gives `1`, it only seems to apply to `/` search, and that's why this code works?

So we do need to turn on `wrapscan`, and set `nomagic` for the EX command to work properly. I think it's more robust than injecting `\M` into the command given that we properly set up try-finally block. But @blueyed mentioned that he [had some issues](https://github.com/junegunn/fzf.vim/pull/140#issuecomment-215572451) with it in #140, I haven't tried it myself. Thoughts?

justinmk:
> excmd holds an EX command in the form of /some-pattern-to-locate-the-position-in-the-file/

If it used `search()` instead of `/` then `\W` flag could be used instead of save/restore of `'wrapscan'` (which adds noise to `:verbose set wrapscan?`).

>  set nomagic | echo 'abc' =~ 'a*' gives 1, it only seems to apply to / search, and that's why this code works?

Sorry if I'm missing context, but why is 'nomagic' needed at all? Why not assume 'magic' and write the patterns as such? `:help 'magic'` notes that no user should change this setting, and fzf.vim is the only plugin I've ever seen set this option. 

The save/restore dance pollutes `:verbose set magic?`.

justinmk:
> Sorry if I'm missing context, but why is 'nomagic' needed at all? Why not assume 'magic' and write the patterns as such?

Oh, I get it now. The patterns are taken from the tags file.

> I think it's more robust than injecting `\M` into the command given that we properly set up try-finally block.

Why is it more robust? Try-finally couldn't hurt, but why not also avoid the save/restore of 'magic'?

justinmk:
At http://ctags.sourceforge.net/FORMAT I see:

```
{tagaddress}	Any Ex command.  When executed, it behaves like 'magic' was
		not set.
```

Oh well. Try-catch it is. 💩

junegunn:
> If it used search() instead of / 

Yeah, the format of excmd part can be different depending on the method used to generate the tags file (`--excmd={number,pattern,mixed}`). It is not guaranteed to be a `/pattern/` and we don't want to care.

blueyed:
Related: https://github.com/universal-ctags/ctags/issues/1125

junegunn:
Fixed in 726d37a6863ac129d4f36ab2f8fc8b9d05bd3971. Thank you.

PR #219, opened 2016-10-13
doc: suggest better default for g:fzf_history_dir

blueyed:
This should be XDG compliant and keeps the user's home dir clean.


junegunn:
Okay, thanks.

PR #222, opened 2016-10-16
[GFiles?] include all untracked files

twe4ked:
By default git-status won't include new files in new directories.

E.g.

    foo/bar/baz.txt # "baz.txt" is a new file in a new directory "bar".

    $ git status --short
    ?? foo/bar/


junegunn:
Looks good, thanks!

twe4ked:
Thanks for the merge.

pavoljuhas:
This is a bit troublesome for repositories that have huge number of uninteresting untracked files.
I for example use git to manage startup files in my HOME directory where this catches tens-of-thousands of irrelevant files under my HOME.

I propose to revert this and use `git config status.showUntrackedFiles all` when untracked files are desired.  See https://github.com/junegunn/fzf.vim/pull/952.

Issue #224, opened 2016-10-19
git status diff fails when running under fish


jsfr:
- Category
    - [ ] Question
    - [x] Bug
    - [ ] Suggestion
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [ ] Vim
    - [x] Neovim

Running under fish gives errors similar to the below when using `:Gfiles?`

```
Illegal command name “(git diff --color=always -- 'server/models/models.js' | sed 1,4d; cat 'server/models/models.js')”
fish: (git diff --color=always -- 'server/models/models.js' | sed 1,4d; cat 'server/models/models.js') | head -500
      ^
```

junegunn:
Because fish is not posix-compliant. I don't know much about fish, is there an alternative command that also works on fish? Or we'll have to juggle with &sh, which I'd like to avoid.

junegunn:
Can you check if this patch helps?

```diff
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index b2aeb47..0610461 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -422,7 +422,7 @@ function! fzf#vim#gitfiles(args, ...)
   let wrapped = fzf#wrap({
   \ 'source':  'git -c color.status=always status --short --untracked-files=all',
   \ 'dir':     root,
-  \ 'options': '--ansi --multi --nth 2..,.. --prompt "GitFiles?> " --preview ''(git diff --color=always -- {-1} | sed 1,4d; cat {-1}) | head -500'''
+  \ 'options': '--ansi --multi --nth 2..,.. --prompt "GitFiles?> " --preview ''sh -c "(git diff --color=always -- {-1} | sed 1,4d; cat {-1}) | head -500"'''
   \})
   call s:remove_layout(wrapped)
   let wrapped.common_sink = remove(wrapped, 'sink*')
```

jsfr:
I unfortunately don't know of a fish alternative to the above command, but the patch fixes it :)

junegunn:
Pushed the fix, thanks.

PR #225, opened 2016-10-22
[Ag] add preview window when ag something

fhfuwei:
By default, ':Ag something' only show the matched line for me.
If we can show the context of the matched line, it will help
us to make decision, like CTRL+SHIFT+F in sublime.

Therefore, I add preview window for Ag command.

screen shot 2016-10-23 at 00 31 54
screen shot 2016-10-23 at 00 33 33

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -505,6 +505,8 @@ endfunction
 " ------------------------------------------------------------------
 " Ag
 " ------------------------------------------------------------------
+let s:fzf_ag_preview = expand('<sfile>:h:h').'/../bin/ag_preview'
+
 function! s:ag_to_qf(line, with_column)
   let parts = split(a:line, ':')
   let text = join(parts[(a:with_column ? 3 : 2):], ':')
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks, but it'll have the same issue as #224, so what do you think about using an external script file located in `fzf.vim/bin/`. That way we can do more sophisticated processing more easily.

fhfuwei:
Good idea!  It's really unreadable if we put the shell script into the vimrc. 
I try to handle this. Thank you.

fhfuwei:
@junegunn I commit the change. Please take look. Thanks

junegunn:
@fhfuwei You're not interested in pushing this further?

fhfuwei:
@junegunn Sorry, I forgot to comment after close. 

I'm interested in this. 

But in my commits, there are hard-code about preview configuration for fzf-tmux.
It's not easy for beginner to customise position of preview pane. I will reopen this pr after refactor code for this. :)

By the way, as you mentioned in https://github.com/junegunn/fzf.vim/issues/224, do you have any plan for other functions? I mean that use "sh -c script" to replace the complicated commands in vimrc. If so, I would like to help you. :P

junegunn:
Most commands are one-liners, so I don't think we need to extract them for now.

By the way, I like your script but would it be possible to extend it to work with syntax highlighting programs like pygmentize, coderay, and highlight?

```sh
# Keep the highlighted input, but decorate line 200 with underline, print around that line to fill the screen
highlight -O ansi README.md | ./focus --line 200
```

fhfuwei:
OK. I got your point.

I think that it is possible for highlight. It is cool thing! 

I will try this.

fhfuwei:
Hi @junegunn 

I tried to highlight text in preview. But it doesn't look well.

`highlight` command: 
1. doesn't support Markdown or Vimrc.
2. I don't know how to highlight the specific line based `highlight` output.

`pygments` command:
It seems that it can support most of file type. But the color scheme looks weird.
I think there are a lot of work to maintain the color setting. 
screen shot 2016-11-02 at 16 26 41

Everyone has different taste, right? 
In my opinion, just let show raw text. :P 
What do you think?

junegunn:
Well, I was not suggesting that we should pick one highlighter but instead implement a text-manipulating filter that can be used with any program, and leave it up to the users to choose which program to use. I'll see if I can implement such a script when I get some time.

By the way, I use `:Ag` frequently for browsing Java projects that has a very long file names and directory structures. And I'm usually on laptops. So I realized that for me, preview window taking up a half of the screen is not always preferable. I'll think about how we should make ag preview customizable and let you know.

fhfuwei:
Cool, Cool, Cool.

I can't wait for the new feature! Haha.

BTW, `stty size < "$(tty)"` is to get size of window. Hope that it can help.

junegunn:
Please try out `grep-preview` branch and let me know what you think.

fhfuwei:
OK.

fhfuwei:
Hi @junegunn, 

To be honest, I cannot figure out better solution. :P 

In my opinion, this feature should be default setting, because someone doesn't know the new feature if the feature is not default setting. 

So, just roll it out~  👍

junegunn:
I think we have a few more things to decide. `--preview-window` value can have `:hidden` suffix which makes fzf start with the preview window hidden. In that case, one has to bind `toggle-preview` action to a key and press the key to make the window appear. But if we only have `g:fzf_grep_preview_window`, it's not easy to set up the binding. Another problem is that the variable is not just used by `Ag` command but also by any custom command based on `fzf#vim#grep` function. And that's why I named the variable `g:fzf_grep_preview_window` instead of `g:fzf_ag_preview_window`, but it may seem strange to the users that only care about `:Ag`.

fhfuwei:
Got your point. Unlike other fzf options, the key binding maybe different for each other. I think we can separate binding option from `opts` and name it `g:fzf_grep_key_binding`. It provides opportunity to us to customize key binding. How about this?

To be honest, I thought that I could use other search command like `ack` when I saw the `fzf#vim#grep` function. HAHA. Back to the key binding issue, I think that it is better to have `g:fzf_grep_command` and `g:fzf_grep_command_options` if `fzf.vim` can support any search tool, like `ack` and `grep`.

fhfuwei:
hi @junegunn 

Sorry to bother you.

What is your next step for the preview feature?  
I just want to know when this feature releases in master. :P

junegunn:
No promises, but I'll try to look into it in a few days. Until then you can use that branch.

fhfuwei:
OK. Got it. Thanks

junegunn:
Pushed different implementation to the master.

https://github.com/junegunn/fzf.vim#advanced-customization

The configuration is a bit more involved, but it allows for more flexibility.

fhfuwei:
👍

PR #230, opened 2016-10-28
[Buffers] Force formatting buffers relative to current dir

andreypopp:
Some plugins sometimes mess up with buffer names and so they become absolute.
This makes searching for buffers with fzf harder. This commit fixes that by
formatting buffers relatively to the cwd so fzf always get relative buffer
names.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -478,7 +478,7 @@ endfunction
 
 function! s:format_buffer(b)
   let name = bufname(a:b)
-  let name = empty(name) ? '[No Name]' : name
+  let name = empty(name) ? '[No Name]' : fnamemodify(name, ":~:.")
   let flag = a:b == bufnr('')  ? s:blue('%', 'Conditional') :
           \ (a:b == bufnr('#') ? s:magenta('#', 'Special') : ' ')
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks.

> Some plugins sometimes mess up with buffer names and so they become absolute.

Can you give me some examples?

andreypopp:
@junegunn wasn't being able to track it down to a specific one but I get this behaviour while writing JS code along with neomake + flowtype.

andreypopp:
I think this should be fixed on their side too... but if you think this change is safe it could be a good safebelt against possible misbehaviours...

andreypopp:
I think I'm hitting this: https://github.com/neovim/neovim/issues/3764 flowtype reports absolute paths which is then parsed by errorformat which makes buffer names become absolute.

junegunn:
I see, but then the comment in the commit is not really accurate. Can you remove those two lines?

andreypopp:
Do you want me to just leave the first one?

andreypopp:
Got you, you want the commit without the comment. Will do.

andreypopp:
I've updated the commit.

junegunn:
Thanks!

PR #239, opened 2016-11-08
Use fugitive (if available) to get the git root

qstrahl:
- Much faster


junegunn:
Thanks!

junegunn:
Hmm, `fugitive#repo().tree()` can set `v:shell_error` to 0 when it's called on an empty buffer.

- cd to a git repo
- vim
- echo fugitive#repo().tree()
- echo v:shell_error

1a2f6c4 addresses the issue.

PR #247, opened 2016-11-18
Make comparison operators user-setting insensitive

Alok:
Changed all operators to strict variants, so user's `ignorecase`
settings don't have to be taken into account.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -53,7 +53,7 @@ endfunction
 function! s:wrap(name, opts, bang)
   " fzf#wrap does not append --expect if sink or sink* is found
   let opts = copy(a:opts)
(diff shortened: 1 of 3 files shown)


junegunn:
I agree that `#` operators are technically safer but I still prefer to use `==` when case-sensitivity doesn't matter, because it looks/reads better. Did you run into any issue with these non-strict operators while using fzf.vim?

Alok:
No, I'm just paranoid about that sort of thing. Looking over your uses of it, it's not really important, so feel free to close this PR.

junegunn:
Okay, I see. [This](https://github.com/junegunn/fzf.vim/pull/247/files#diff-7274b7c6c87526542ef9087f2e997dfaR714) makes sense though, I'll apply the change and close the PR. Thanks.

PR #252, opened 2016-11-26
ctags redirects 2>/dev/null

dprophete:
This is a fairly tiny change.
It simply redirects the ctags error to /dev/null in case the current project contains some files which create errors with ctags.


junegunn:
Merged, thanks.

dprophete:
Thanks @junegunn . That's was super fast.

PR #253, opened 2016-11-26
use absolute path for windows and buffers

dprophete:
Currently, displaying the list of windows/buffers will use the path for the current edited file as the base path for all other paths. The means the list of paths will differ based on the path of the current file. 

This PR simply makes sure we 'cd /' first and then replaces the $HOME part of the path by '~'

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -507,6 +507,7 @@ function! s:sort_buffers(...)
 endfunction
 
 function! fzf#vim#buffers(...)
+  cd /
   let bufs = map(sort(s:buflisted(), 's:sort_buffers'), 's:format_buffer(v:val)')
   return s:fzf('buffers', {
   \ 'source':  reverse(bufs),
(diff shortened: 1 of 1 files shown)


junegunn:
> The means the list of paths will differ based on the path of the current file.

It's not based on the current file, but on the current directory. Maybe are you using `autochdir`? Anyway, I prefer the current behavior, since I usually stays on the project root, so relative paths are more concise and straightforward. Also we should try to avoid changing the global state, in this case `cd /`).

dprophete:
Indeed. I am using `autochdir` (I do like being able to do ':e <tab>' and see the files in the same dir as the file I am currently editing), but as you mentioned not everybody uses it.

Ok, so maybe I can only `cd /` if the `autochdir` option is set. And then I'll revert back to the previous dir, so as not change the global state.

junegunn:
Consider setting up a custom command that wraps Buffers or Files command to implement the idea.

PR #254, opened 2016-11-26
Allow restyling of the fzf[123] groups

Soares:
Tiny change. Allows users to set the styling on the fzf1, fzf2, and fzf3 highlight groups.


junegunn:
Merged, thanks.

PR #255, opened 2016-11-26
Windows and buffer paths relative to root

dprophete:
Let's try this again :-)

This modifies the base `fzf#vim#buffers` and `fzf#vim#windows` function to displays the files relative to the root directory, only when autochdir is set. 

It does this by introducing a generic wrapper function `s:execute_in_root_dir`.


junegunn:
Thanks for the suggestion. However, setting `&autochdir` does not necessarily mean that the user prefers absolute paths. We could probably add an option so that the user can choose the behavior. Or like I said in the previous comment, one can define a custom command that implements the idea without changing fzf.vim. I believe this approach is simpler and more flexible (one can easily switch between both behaviors just by using different commands).

```vim
command! B cd / | Buffers | cd -
command! W cd / | Windows | cd -
```

PR #259, opened 2016-12-03
add preview.sh for ruby-less previews.

acornejo:
I don't use ruby, but would still like to enjoy the fzf preview feature.
This adds a preview.sh script as a bare-bones alternative to preview.rb
for those without ruby installed.

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -26,18 +26,19 @@ if [ -z "$CENTER" ]; then
     CENTER=1
 fi
 
-LINES=40
-COLUMNS=80
-
 if [ -r /dev/tty ]; then
-    SIZE=$(stty size < /dev/tty)
-    LINES=$(echo $SIZE | awk '{print $1}')
-    COLUMNS=$(echo $SIZE | awk '{print $2}')
+    LINES=$(stty size < /dev/tty | awk '{print $1}')
+else
+    LINES=40
 fi
 
+if [ -n "$SPLIT" ]; then
+    LINES=$(($LINES/2)) # using horizontal split
+fi
 LINES=$(($LINES-2)) # remove preview border
-FIRST=$(($CENTER-$LINES/2))
-FIRST=$(($FIRST < 1 ? 1 : $FIRST))
-LAST=$((${FIRST}+${LINES}))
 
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks, I compared the output of the two scripts side by side and noticed a couple differences. 

screen shot 2016-12-04 at 12 58 37 am

Please see my comments.

acornejo:
Thanks for taking the time to review this. I don't see any comments though, but here is the list of differences I observe (btw I can't run preview.rb so can't compare side-by-side myself, so this is useful).

- preview.sh centers the highlighted line, while preview.rb shows it at 1/3rd of the screen (why btw?)
- preview.sh does not use highlight or coderay. I assumed these were ruby tools as well, so I skipped them, but I'll add checks to see if they are present and use them when available.
- preview.sh truncates after 60 characters. I did this intentionally to prevent line wrapping, since I couldn't figure out a way to determine the screen width (I filed an issue in fzf to avoid line wrapping on preview windows).

Does that make sense?

junegunn:
I forgot to submit the review. The comments should be visible now.

> while preview.rb shows it at 1/3rd of the screen (why btw?)

1. Because I felt it was easier to get the context that way. It's subjective though. 2. And partially because of possible wrappings of the preceding lines. It's going to be a non-issue when nowrap option gets in.

bin/preview.sh
```diff
@@ -0,0 +1,43 @@
...
+COLUMNS=80
+
+if [ -r /dev/tty ]; then
+    SIZE=$(stty size < /dev/tty)
+    LINES=$(echo $SIZE | awk '{print $1}')
+    COLUMNS=$(echo $SIZE | awk '{print $2}')
+fi
+
+LINES=$(($LINES-2)) # remove preview border
+FIRST=$(($CENTER-$LINES/2))
```
acornejo: done

acornejo:
made changes.

Still not using coderay/highlight since I realized getting the reversed line effect with text that is already ANSI highlighted is not trivial to do in awk. It would be trivial to do in gawk (gnu extensions) but unfortunately most systems (including linux and osx) don't have that installed by default.

junegunn:
Looks good, thanks!

PR #278, opened 2016-12-24
[fzf#vim#colors] Remove duplicate colorschemes

mklein994:
This makes it work more intuitively with plugins like [flazz/vim-colorschemes](https://github.com/flazz/vim-colorschemes).

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -358,7 +358,7 @@ endfunction
 " ------------------------------------------------------------------
 function! fzf#vim#colors(...)
   return s:fzf('colors', {
-  \ 'source':  uniq(sort(map(split(globpath(&rtp, "colors/*.vim"), "\n"),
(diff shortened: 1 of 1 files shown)


junegunn:
Please update the code to use `s:uniq()`. `uniq()` is not available on older versions of Vim.

mklein994:
This would be my first venture into VimL, as I'm fairly new to writing plugins for Vim, but I could give it a shot. Do you have any style guidelines you want me to follow?

junegunn:
There's already `s:uniq` function defined in that file. All you have to do is to replace `uniq (...)` with `s:uniq (...)`.

mklein994:
Can't believe I didn't see that… ah well.

junegunn:
Thanks for the update. Oh, by the way, can you remove `sort()`? `s:uniq` does not require the list to be sorted. Currently, non-system color schemes installed with vim-plug are listed before the ones that come with Vim by default, and I like the ordering.

junegunn:
Merged, thanks!

PR #284, opened 2017-01-06
fix :Tags by adding -a option to grep

### LINKED ISSUE #5: Problem with :Tags
The only output I get is `Binary file (standard input) matches`.

I have hunted the issue to [this line](https://github.com/junegunn/fzf.vim/blob/master/plugin/fzf.vim#L382).

I get the same issue when I run it in my terminal though, so it might be a perl issue.

andihit:
Somehow the tags file generated by ctags for the linux kernel, tag
v4.10-rc1 contains non-ASCII characters. grep stops when it detects
non-ASCII characters. This patch adds the -a option to the grep command
to treat the tags file as ASCII text.

Steps to reproduce:
```
$ git clone git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git
$ cd linux
$ git checkout v4.10-rc1
$ make O=. SRCARCH=x86 tags
$ grep -v '^!' tags | tail
Binary file tags matches
```

Another possibility is to call grep with `LC_ALL=C`.

(Fixes #5)

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -766,7 +766,7 @@ function! fzf#vim#tags(query, ...)
   " We don't want to apply --ansi option when tags file is large as it makes
   " processing much slower.
   if getfsize(tagfile) > 1024 * 1024 * 20
-    let proc = 'grep -v ''^\!'' '
+    let proc = 'grep -av ''^\!'' '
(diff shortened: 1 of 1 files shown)


junegunn:
Merged, thanks.

PR #306, opened 2017-02-08
when checking exists of commands it checks full match

sangwook:
if i defined commands start with `Ag`. (e.g. [1]) fzf.vim can't define `:Ag` command.
[1]: `command! AgGem call fzf#vim#ag(<q-args>)`


junegunn:
Thanks, I didn't know that it returns non-zero value for partial matches 😮

PR #332, opened 2017-03-08
Support tagrelative in Tags

pokey:
This patch fixes `:Tags` to respect the `tagrelative` setting when resolving paths in the tagfile.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -780,11 +780,17 @@ function! fzf#vim#tags(query, ...)
     let proc = 'perl -ne ''unless (/^\!/) { s/^(.*?)\t(.*?)\t/'.s:yellow('\1', 'Function').'\t'.s:blue('\2', 'String').'\t/; print }'' '
     let copt = '--ansi '
   endif
-  return s:fzf('tags', {
-  \ 'source':  proc.shellescape(fnamemodify(tagfile, ':t')),
+  let opts = {
   \ 'sink*':   s:function('s:tags_sink'),
-  \ 'dir':     fnamemodify(tagfile, ':h'),
-  \ 'options': copt.'-m --tiebreak=begin --prompt "Tags> "'.s:q(a:query)}, a:000)
+  \ 'options': copt.'-m --tiebreak=begin --prompt "Tags> "'.s:q(a:query)
+  \}
+  if (&tagrelative)
+     let opts['dir'] = fnamemodify(tagfile, ':h')
+     let opts['source'] = proc.shellescape(fnamemodify(tagfile, ':t'))
+  else
+     let opts['source'] = proc.shellescape(tagfile)
+  endif
+  return s:fzf('tags', opts, a:000)
 endfunction
 
 " ------------------------------------------------------------------


junegunn:
Can you elaborate more on the problem you have with `tagrelative`? `:Tags` command works for me with or without the option for me. It would be nice if you could present simple steps to reproduce the problem.

pokey:
You need to have your tags file in a different directory from cwd / root of repo but have paths relative to root of repo. I keep my tags file at .git/tags but paths are relative to root of repo.  So you should be able to reproduce this by generating your ctags at base of repo and then moving them into .git.  Ie run this at the base of a repo:

```
ag -l | ctags -L -
mv tags .git
```

Then run vim at base of repo and let fugitive find the tags in .git, as mentioned [here](http://tbaggery.com/2011/08/08/effortless-ctags-with-git.html).  With `set notagrelative`, vim is able to resolve paths in the tags file correctly, but `:Tags` is not.  

Fwiw I'm leaning towards just moving my tags from `.git/tags` to `.tags` and setting `tagrelative` again, as I don't think it's worth the headache, but I figure since vim supports `notagrelative`, fzf.vim might as well support it too.

pokey:
The way the plugin would fail is that it would cd into the parent directory
of the tags file before resolving the paths in the tags file. However, the
paths in my tags file are supposed to be relative to the cwd, which I
always have as the root of the repo. So these paths wouldn't resolve
correctly when fzf.vim tried to open them. It would load an empty buffer.
Interestingly, running :e on this buffer then loads the file correctly.

Your code mirrors the semantics of tagrelative, but the semantics of
notagrelative dictate that the paths be resolved relative to cwd.

junegunn:
I see, that makes sense. By the way, I'm currently more interested in supporting multiple tags files (i.e. `tagfiles()` returning multiple files), which is useful for locating stuff in the dependent libraries of the current project. I presume `notagrelative` will not play nicely with multiple tags file setup?

pokey:
I haven't tested this, but so long as the paths in each tag file are
relative to the cwd I don't see why multiple tag files would be a problem
with `notagrelative`. I think that would be impossible to ensure if the
libraries are imported by multiple projects, unless you just used absolute
paths in those tag files to avoid the problem altogether, but in principle
it should work.

All this being said, between the last time I wrote and now, I've stopped
using `notagrelative` and just put a .tags file at the base of the repo, as
it seemed too fragile not to have the paths relative to the tag file, and I
couldn't stand having `../` at the beginning of all the paths.

But either way I don't see the harm in having fzf.vim support
`notagrelative` in case someone decides to use it for some reason.

junegunn:
> I haven't tested this, but so long as the paths in each tag file are relative to the cwd

I think it's practically not possible, since tags files are generated individually for each library, and each library can be used in multiple projects in different paths (For example, I use [bundler.vim](https://github.com/tpope/vim-bundler) which automatically brings in tags files of all dependent gems). Using `notagrelative` means that you have to recreate tags files for all the dependent libraries whenever you switch between your projects.

pokey:
Yes that's what I was trying to say with my second sentence, but I guess I
could have been a bit more clear :-). And yeah for this and a host of other
reasons I've decided that `notagrelative` is not really a good idea.

junegunn:
Yeah, I'm going to pass on this one, thanks for the input. I updated Tags command to work with multiple tags files.

pokey:
Hmm ok.  I stopped using `notagrelative` so doesn't make a difference in my
setup anymore.  Might want to at least make a note somewhere that
`notagrelative` isn't supported.  But guessing it's not a widely used
feature so probably not a big deal

PR #363, opened 2017-05-01
Add preview to `History`

Alok:
That way it's easy to tell which old files we've been reading.


junegunn:
Thanks, but that's not how we do it. You can use autoload functions like follows.

```vim
command! -bang Oldfiles call fzf#vim#history(fzf#vim#with_preview(), <bang>0)
```

Refer to https://github.com/junegunn/fzf.vim#advanced-customization to learn more about `fzf#vim#with_preview`.

Alok:
In that case, can you consider using that wrapper for `History` by default?

junegunn:
Not everyone likes preview window but I'll think about it.

PR #383, opened 2017-06-04
Add HgFiles command to run fzf on Mercurial repos.

voithos:
This uses `hg status` along with the relevant flags to only display existing, non-hgignore'd files (and new, unknown files).


junegunn:
Thanks, but I'm not interested in adding more commands to this repository at the moment due to maintenance burden. We can maybe add the command to the wiki page https://github.com/junegunn/fzf/wiki/Examples-(vim)

voithos:
OK. It'd be nice if the README had a `Contributing` section, or something along those lines, to prevent others from doing the same thing needlessly.

junegunn:
Good point, thanks.

PR #372, opened 2017-05-13
Support Windows (requires 0.16.10)

janlazo:
Main PR to fix commands for Windows and use list type for options if required.
Added workarounds for commands not using list type for options such as `:Helptags`

`fzf#shellescape` is not used in this PR so `shellslash` is set/unset depending on the function.
- `set shellslash`
  - path functions such as `expand` and `glob` to avoid `\"`
  - `shellescape` commands on sh
- `set noshellslash`
  - `shellescape` commands on cmd.exe

**Requirements**
-  fzf [0.16.10](https://github.com/junegunn/fzf-bin/releases/tag/0.16.10) or later versions so commands generating relative filepaths will work in both vim/neovim (junegunn/fzf#969 and junegunn/fzf#970)
- Cygwin utilities (sh, grep, perl, etc.) available in `PATH` to reduce maintenance burden
- Windows 7 or later versions
- Vim 7.4+ or Neovim 0.2+

**Functions**
- [ ] `s:fzf` (48ad1c3, supports list type for options)
- [ ] `s:wrap` (48ad1c3, supports list type for options)
- [X] `fzf#vim#grep` (9a72951)
  - requires list type for option if prompt contains quotes or escape characters (`\^"`)
- [ ] `fzf#vim#complete`
  - [X] `s:complete_trigger` (da8f862)
  - [ ] `fzf#vim#complete#word` (f1619d4, not supported)
    - relies on `/usr/share/dict/words` which does not exist in Windows (no equivalent??)
  - [ ] `fzf#vim#complete#path` (24c6594)
    - requires `set shellslash` for now
    - TODO: use list type for options to resolve shellescape issues in cmd.exe
  - [ ] `fzf#vim#complete#line`
  - [ ] `fzf#vim#complete#buffer_line`
- [X] `fzf#vim#map` (no fixes required)
- [X] `fzf#vim#with_preview` (9535460)

**Commands**
- [X] `Files` (48ad1c3)
- [x] `GFiles` (no fixes required)
- [x] `GFiles?` (7e5a6af, list type for options not required, requires sh.exe from Cygwin)
- [X] `Buffers` (48ad1c3)
- [x] `Colors` (no fixes required)
- [x] `Ag` (a943f21)
  - depends on `fzf#vim#grep`
- [X] `Lines` (cb86ba6)
- [X] `BLines` (48ad1c3)
- [X] `Tags` (7ed4abb)
- [X] `BTags` (7cfbeca)
- [X] `Marks` (no fixes required)
[...]

PR #386, opened 2017-06-08
Use fnamemodify in Cygwin for Windows paths

janlazo:
From initial testing, only `Files` and `GFiles` require fnamemodify.

I can't test `Locate` in cygwin because it requires updatedb and locate needs a regularly updated database. Windows has its own indexing for explorer.exe but I disabled it in my system for performance. I'll revist it in #372

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -246,6 +246,9 @@ function! fzf#vim#files(dir, ...)
       return s:warn('Invalid directory')
     endif
     let dir = substitute(a:dir, '/*$', '/', '')
+    if has('win32unix')
+      let dir = fnamemodify(dir, ':p')
+    endif
     let args.dir = dir
     let args.options .= ' --prompt '.shellescape(dir)
   else
@@ -463,6 +466,9 @@ function! fzf#vim#gitfiles(args, ...)
   let root = s:get_git_root()
   if empty(root)
     return s:warn('Not in git repo')
+  endif
+  if has('win32unix')
+    let root = fnamemodify(root, ':p')
   endif
   if a:args != '?'
     return s:fzf('gfiles', {


junegunn:
Thanks, this makes me wonder, would it make sense to do `:p` expansion inside `fzf#run`?

janlazo:
Sure, since it simplifies `FZF` and `s:get_git_root` does not require `fnamemodify` for `Commits`.

I see `dict.dir` handled in [fzf#run](
https://github.com/junegunn/fzf/blob/7e483b0c25224ae218bcc2fb67896d6b4d1c8ea0/plugin/fzf.vim#L361-L363)
I don't know why there's a `has('nvim')` condition.

janlazo:
Need `fnamemodify` only if `dict.dir` exists.
`getcwd` always returns a unix-style path in cygwin vim.

junegunn:
It's only for the cases where `dir` is not given and it's not related to the current issue. You can always `git blame` on that :)

https://github.com/junegunn/fzf/commit/cd59e5d07b3a69a75f7c0d02efd66b6f04bb13a1

Let's just make sure that we're on the same page. `:p` expansion is only needed on Cygwin to handle network drives, and the following patch on the main repo should do, right?

```vim
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 99fda1f..64e8cd3 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -360,6 +360,8 @@ try
 
   if has('nvim') && !has_key(dict, 'dir')
     let dict.dir = s:fzf_getcwd()
+  elseif has('win32unix')
+    let dict.dir = fnamemodify(dir, ':p')
   endif
 
   if !has_key(dict, 'source') && !empty($FZF_DEFAULT_COMMAND)
@@ -765,8 +767,6 @@ function! s:cmd(bang, ...) abort
     let opts.dir = substitute(substitute(remove(args, -1), '\\\(["'']\)', '\1', 'g'), '[/\\]*$', '/', '')
     if s:is_win && !&shellslash
       let opts.dir = substitute(opts.dir, '/', '\\', 'g')
-    elseif has('win32unix')
-      let opts.dir = fnamemodify(opts.dir, ':p')
     endif
     let prompt = opts.dir
   else
```

junegunn:
By the way, I wonder if neovim runs on cygwin.

junegunn:
fixup.

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 99fda1f..bad7449 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -361,6 +361,9 @@ try
   if has('nvim') && !has_key(dict, 'dir')
     let dict.dir = s:fzf_getcwd()
   endif
+  if has('win32unix') && has_key(dict, 'dir')
+    let dict.dir = fnamemodify(dict.dir, ':p')
+  endif
 
   if !has_key(dict, 'source') && !empty($FZF_DEFAULT_COMMAND)
     let temps.source = s:fzf_tempname().(s:is_win ? '.bat' : '')
@@ -765,8 +768,6 @@ function! s:cmd(bang, ...) abort
     let opts.dir = substitute(substitute(remove(args, -1), '\\\(["'']\)', '\1', 'g'), '[/\\]*$', '/', '')
     if s:is_win && !&shellslash
       let opts.dir = substitute(opts.dir, '/', '\\', 'g')
-    elseif has('win32unix')
-      let opts.dir = fnamemodify(opts.dir, ':p')
     endif
     let prompt = opts.dir
   else
```

janlazo:
Thanks for the reminder. I use vim-rooter so thanks for the patch :)
I'm fine with the patch but need to test the prompt.

> By the way, I wonder if neovim runs on cygwin.
It does if you don't use mintty.

janlazo:
The prompt works for me and is better to me because if I pass a Windows-style path to `FZF`, I expect the same style in the prompt.

junegunn:
Okay, thanks for the feedback. Let me push the fix.

janlazo:
Thanks.

Issue #410, opened 2017-07-29
Feature: Support list type for fzf options


janlazo:
- Category
    - [ ] Question
    - [ ] Bug
    - [X] Suggestion
- OS
    - [ ] Linux
    - [ ] macOS
    - [X] Windows
    - [ ] Etc.
- Vim
    - [X] Vim
    - [X] Neovim

#372 works in Windows but it is messy and I did not consider user settings and combinations of commands (such as preview with ag). Since Windows support requires non-trivial shellescaping via list type for fzf options, it is better to support it regardless of the OS by starting off with `Files`, `s:fzf`, and `s:wrap`.  `FZF` works in Windows/Cygwin on Vim/Neovim and handles non-trivial shellescaping in cmd.exe.

The goal is to resolve all issues around `Files` so that is easier to support Windows without breaking backwards compatibility.

janlazo:
Status of commands and functions in fzf.vim regarding this ticket :

**Functions**
- [X] `s:fzf`
- [X]  `s:wrap`
- [ ] `fzf#vim#preview`
  - (Windows) disabled because preview scripts in bin/ do not work with the default command, `dir /s/b`

**Commands**
- [X] ``Files``
- [X] `GFiles?`
  - (Windows) requires Unix utilities in PATH
- [X] `Buffers`
- [X] `Lines`
- [X] `BLines`
- [X] `[Helptags]`

janlazo:
All required PRs for Windows support are merged and `fzf#shellescape` handles all escaping issues if using list type for options.
List type for options is optional.
Unless I missed a command or user option, I consider this issue resolved.

PR #418, opened 2017-08-14
[Files] runs in Windows

janlazo:
Initial work for  #410.
I added `s:merge_opts` to handle `eopts` so it can fallback to string options.
I didn't use `fzf#shellescape` so it work in Window only if `g:fzf_files_options` is unset or a list.

Maybe it's better to not support string values for these user options in Windows.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -55,9 +55,6 @@ endfunction
 function! fzf#vim#with_preview(...)
   " Default options
   let options = {}
-  if s:is_win
-    return options
-  endif
   let window = 'right'
 
   let args = copy(a:000)
@@ -66,6 +63,9 @@ function! fzf#vim#with_preview(...)
   if len(args) && type(args[0]) == s:TYPE.dict
     let options = copy(args[0])
     call remove(args, 0)
+  endif
+  if s:is_win
+    return options
   endif
 
   " Preview window
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -28,12 +28,46 @@ set cpo&vim
 " Common
 " ------------------------------------------------------------------
 
+let s:is_win = has('win32') || has('win64')
+if s:is_win
+  function! s:fzf_call(fn, ...)
+    let shellslash = &shellslash
+    try
+      set noshellslash
+      return call(a:fn, a:000)
+    finally
+      let &shellslash = shellslash
+    endtry
+  endfunction
+else
+  function! s:fzf_call(fn, ...)
+    return call(a:fn, a:000)
+  endfunction
+endif
+
+function! s:fzf_shellescape(str)
+  let escaped =  s:fzf_call('shellescape', a:str)
+  return s:is_win ? substitute(escaped, '[^\\]\zs\\$', '\\\\', '') : escaped
+endfunction
+
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:bin_dir = expand('<sfile>:h:h:h').'/bin/'
 let s:bin = {
 \ 'preview': s:bin_dir.(executable('ruby') ? 'preview.rb' : 'preview.sh'),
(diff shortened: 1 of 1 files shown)


junegunn:
> I didn't use fzf#shellescape so it work in Window only if g:fzf_files_options is unset or a list.

Sorry I don't follow, can you elaborate? It seems to me that we can still use it in line 68 for building options string.

janlazo:
Based on https://github.com/junegunn/fzf/pull/916#issuecomment-303621871, I thought I shouldn't use `fzf#shellescape` in fzf.vim so I avoided it in here and #372

junegunn:
I see, you're right, that was my initial thought. But since it's unlikely that we're going to remove it in the future, let's just use `fzf#shellescape` for now and avoid code duplication.

Also, you probably should `copy(v:val) in `join(map(v:val ...`.

janlazo:
Any other function or option used with `Files`?

janlazo:
Should `Files` work with `fzf#vim#with_preview`?

junegunn:
It does.

```vim
call fzf#vim#files('.', fzf#vim#with_preview())
```

So this is about `with_preview` only working with string options, right?

```vim
" This works
:echo fzf#vim#with_preview({'options': '--reverse'})

" But this currently doesn't. It should be fixed.
:echo fzf#vim#with_preview({'options': ['--reverse']})
```

janlazo:
I get `File not found: C` for the preview.

janlazo:
It doesn't work with the bash script in Windows.

junegunn:
I see, thanks for testing it. So, two things:

1. `fzf#vim#preview` should be able to handle list options (whether you're on Windows or not)
2. Make `fzf#vim#preview` no-op (return argument as it is) on Windows at least for now

janlazo:
```diff
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index 9f10e44..1289f4d 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -74,12 +74,14 @@ function! fzf#vim#with_preview(...)
     call remove(args, 0)
   endif

-  let preview = ['--preview-window', window, '--preview', s:bin.preview.' '.(window =~ 'up\|down' ? '-v' : '').' {}']
+  if !s:is_win
+    let preview = ['--preview-window', window, '--preview', s:bin.preview.' '.(window =~ 'up\|down' ? '-v' : '').' {}']

-  if len(args)
-    call extend(preview, ['--bind', join(map(args, 'v:val.":toggle-preview"'), ',')])
+    if len(args)
+      call extend(preview, ['--bind', join(map(args, 'v:val.":toggle-preview"'), ',')])
+    endif
+    call s:merge_opts(options, preview)
   endif
-  call s:merge_opts(options, preview)
   return options
 endfunction
```

junegunn:
Yes, I would return just it at the beginning of the function

```vim
if s:is_win
  return options
endif

...
```

autoload/fzf/vim.vim
```diff
@@ -55,6 +55,9 @@ endfunction
 function! fzf#vim#with_preview(...)
   " Default options
   let options = {}
+  if s:is_win
+    return options
+  endif
```
junegunn: Sorry for the confusion, this should come after line 69.

janlazo:
Using `set shellslash` with `expand` is a quick workaround for the preview script to run.
The network drive is remaining issue because of the colon.

janlazo:
I prefer for the default command to output relative paths (see https://github.com/junegunn/fzf/pull/971#issuecomment-313971408) because I can't see the tail of the filepath in the cmd.exe terminal when running the preview.

It will take awhile for `fzf#vim#with_preview` to work in Windows so `no-op` is the only option

junegunn:
Is this ready for merge?

janlazo:
Let's merge for now since there's no other interaction I can think of with `Files`.

junegunn:
Merged, thanks!

PR #422, opened 2017-08-17
exclude the current file from History

oconnor663:
This is the first time I've ever written VimL, so I'm sure there's a better way to do it. But this seems to achieve what we were looking for in https://github.com/junegunn/fzf.vim/issues/367 :) @junegunn do you think excluding the current file would be a good idea in general?


junegunn:
I think it makes sense because we already do so with `:Buffers` (which I personally use far more often than `:History`). I'll merge this, thanks.

PR #423, opened 2017-08-19
[Helptags] works in Windows

janlazo:
Moved the perl script to a temporary file (don't know when to delete it) to avoid escaping issues.
Not sure about the filepath regex to resolve the network drive but it does the job

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -941,12 +941,6 @@ function! s:helptag_sink(line)
   execute 'help' tag
 endfunction
 
-if exists('s:helptags_script')
-  silent! call delete(s:helptags_script)
-endif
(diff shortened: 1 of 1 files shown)


junegunn:
How about storing the path as a script local variable and deletes if it's defined? So that temp files are not piling up while Vim is up.

```vim
if exists('s:helptags_script')
  silent! call delete(s:helptags_script)
endif

let s:helptags_script = ...

janlazo:
What about deleting temp files on `VimLeavePre`?

junegunn:
Oh, I was not suggesting that we extract it out of the function. No need to create the file when even before `:Helptags` is used. I wouldn't worry too much about not cleaning up tempfile when leaving Vim if there aren't too many.

PR #424, opened 2017-08-19
[GFiles?] works in Windows

janlazo:
Initial work for `GFiles?` in Windows.
The preview command hasn't changed but Windows doesn't have the required tools by default.


junegunn:
Thanks!

PR #425, opened 2017-08-20
[Lines,BLines,Buffers] works in Windows

janlazo:
Need https://github.com/junegunn/fzf/pull/1021 for the backslash queries
`s:q` uses `fzf#shellescape` now but `Tags` and `BTags` use `s:q`.

What do you think of removing/reworking `s:q` and use list type for options?


junegunn:
> What do you think of removing/reworking s:q and use list type for options?

No objections. I'll merge this for now, we can gradually migrate to list options.

PR #426, opened 2017-08-20
If `/` remapped, `History/` command will block vim.

### LINKED ISSUE #409: `FZFHistory/` doesn't work.
- Category
    - [ ] Question
    - [x] Bug
    - [ ] Suggestion
- OS
    - [ ] Linux
    - [x] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [x] Vim
    - [ ] Neovim

minimal configuration: https://gist.github.com/tenfyzhong/352983d4de9d50640c668961b43ba4d8
When I use `FZFHistory/` to history search. The vim is blocked. 
Because I use [haya14busa/incsearch.vim](https://github.com/haya14busa/incsearch.vim) and set a mapping,  `map / <Plug>(incsearch-forward)`. 
I think this line: https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L408 
should change to `call feedkeys(a:type.item."\<cr>", 'n')`, it works for me.

tenfyzhong:
fixes #409

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -405,7 +405,9 @@ function! s:history_sink(type, lines)
     call feedkeys(a:type."\<up>")
   else
     let g:__fzf_command = "normal ".a:type.item."\<cr>"
-    call feedkeys("\<plug>(-fzf-vim-do)")
+    " add by tenfy(tenfyzhong@qq.com) 
+    " If `/` remapped, this called will block vim. 
+    call feedkeys(a:type.item."\<cr>", 'n')
   endif
 endfunction
 


junegunn:
I think this is an issue of incsearch plugin you use, it's causing `execute "normal /darwin\<cr>"` to block when it's a perfectly valid command. You might want to report the issue to the author of the plugin.

We can make `History/` bypass `n` mapping by changing `normal` to `normal!` (or using `feedkeys` with `n` argument as you suggested), but I won't do it, because there can be some other plugins that would do some extra bookkeeping via its own `n` mapping.

tenfyzhong:
The `History/` call the `/` search. You don't know what the user map `/` to, so you should make sure the `History/` can work.

junegunn:
Yes, you have a point. But I use a plugin that remaps `/` and I want it to be triggered on `:History/`. As you can guess from what I mentioned on README page, this plugin was built around my own preferences, and I'm not sure if I want to change it for a plugin I don't use. Is it really not possible to fix the issue of the plugin?

tenfyzhong:
Can you export the function `s:fzf`( https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L184) as `fzf#vim#fzf`?In this way, I can define the history search by myself and it's a good way for user to define something he want to search.

junegunn:
For that purpose, you should use `fzf#run` and `fzf#wrap` from the main repo.

https://github.com/junegunn/fzf/blob/master/README-VIM.md#fzfrun

fzf.vim project is just a collection of functions and commands built from those two to demonstrate what can be done with fzf on Vim.

PR #427, opened 2017-08-20
[Tags, BTags] works in Windows

janlazo:
It works with relative paths only.
`s:open` is not using `fnameescape`.


junegunn:
Looks good to me, thanks!

janlazo:
I forgot to fix the sink

junegunn:
I see, would you send me another PR for that?

PR #429, opened 2017-08-20
[Ag, fzf#vim#grep] works in Windows

janlazo:
It should work with the preview because ag does not output the network drive but the absolute filepath of the preview script must be modified with `fnamemodify` and  either `set shellslash` or `escape(abspath, '\')` to get around escaping issues.

Ideally, `s:fzf` should return early if running on windows, the preview script is used and if there is either no source or `FZF_DEFAULT_COMMAND` is undefined such that fzf defaults to `dir /s/b`.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -205,7 +205,8 @@ function! s:fzf(name, opts, extra)
   let eopts  = has_key(extra, 'options') ? remove(extra, 'options') : ''
   let merged = extend(copy(a:opts), extra)
   call s:merge_opts(merged, eopts)
-  if s:is_win && empty(get(merged, 'source', '')) && empty($FZF_DEFAULT_COMMAND) && get(merged, 'options', '') =~# s:bin.preview
+  let mopts = get(merged, 'options', '')
+  if s:is_win && empty(get(merged, 'source', '')) && empty($FZF_DEFAULT_COMMAND) && (type(mopts) == s:TYPE.list ? join(mopts) : mopts) =~# s:bin.preview
     return s:warn('preview script is incompatible with the default command in Windows')
   endif
   return fzf#run(s:wrap(a:name, merged, bang))
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -35,6 +35,15 @@ let s:bin = {
 \ 'preview': s:bin_dir.(executable('ruby') ? 'preview.rb' : 'preview.sh'),
 \ 'tags':    s:bin_dir.'tags.pl' }
 let s:TYPE = {'dict': type({}), 'funcref': type(function('call')), 'string': type(''), 'list': type([])}
+if s:is_win
+  if &shellslash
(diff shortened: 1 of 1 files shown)


janlazo:
The fnamemodify hack doesn't work in Neovim.
I have to escape with double quotes otherwise filepaths with spaces will break the preview command

janlazo:
I'm keeping the preview script unescaped for now.
I can't think of a robust solution without resorting to a batchfile for proper escaping.

junegunn:
> Ideally, s:fzf should return early if running on windows, the preview script is used and if there is either no source or FZF_DEFAULT_COMMAND is undefined such that fzf defaults to dir /s/b.

But there's no guarantee that a custom FZF_DEFAULT_COMMAND or source command is compatible with the preview script, right?

janlazo:
Yes, but `fzf#vim#with_preview` is optional. There are several ways to generate relative filepaths, including regular grep.

However, I don't think this is an issue now. `:` is not allowed in Windows filepaths anyway so resolving the network drive is detecting the regex `^[A-Z]:\\`, splitting the filepath with `:` delimiter, and performing `cd /d %network_drive%` before running the preview script.

This can be resolved via a wrapper batchfile or in the ruby/bash scripts.

autoload/fzf/vim.vim
```diff
@@ -35,6 +35,14 @@ let s:bin = {
 \ 'preview': s:bin_dir.(executable('ruby') ? 'preview.rb' : 'preview.sh'),
 \ 'tags':    s:bin_dir.'tags.pl' }
 let s:TYPE = {'dict': type({}), 'funcref': type(function('call')), 'string': type(''), 'list': type([])}
+if s:is_win
+  if has('nvim')
+    let s:bin.preview = split(system('for %A in ("'.s:bin.preview.'") do echo %~sA'), "\n")[1]
```
junegunn: Can you explain how this line works? Any documentation I can refer to?

junegunn:
Merged, thanks a lot!

janlazo:
I'll send a new PR to resolve absolute paths for the preview command via a batchfile.
Unless you want to handle this absolute path issue in the fzf binary directly

autoload/fzf/vim.vim
janlazo: I used a simple variant of the `FOR` command in cmd.exe and parameter extension on the for loop variable. It runs the command after `do` so [0] matches `echo %~sA` and `[1]` matches the output of that `ECHO` command. The [1] output is an absolute path with backslash for path separator so the `escape` is necessary for fzf binary to not remove the backslashes when it runs the preview command.

Aside from different path separators because of `set shellslash`, `fnamemodify` and that line should output the same absolute path.

Documentation:
https://ss64.com/nt/for2.html
https://ss64.com/nt/syntax-args.html

janlazo: There is a low risk of 8.3 filename disabled manually (by administrators) but the common case is it remains enabled to support legacy software.

PR #430, opened 2017-08-21
[Commits,BCommits] works in Windows

janlazo:
Works only in `s:execute` of fzf's Vim plugin.
`s:execute_term` corrupts the output of `fzf#shellescape`.


junegunn:
Looks good to me, thanks.

PR #434, opened 2017-08-23
make :History show recent files first, and filter duplicates

oconnor663:
Also normalize paths with :~:. These changes make the output of :History
more similar to CtrlP's MRU mode.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -407,10 +407,38 @@ endfunction
 " ------------------------------------------------------------------
 " History[:/]
 " ------------------------------------------------------------------
-function! s:all_files()
-  return extend(
-  \ filter(reverse(copy(v:oldfiles)), "filereadable(expand(v:val))"),
-  \ filter(map(s:buflisted(), 'bufname(v:val)'), '!empty(v:val)'))
+function! s:history_files()
+  " Use a dictionary to detect duplicates, between the buffers list and the
+  " oldfiles list. Prepopulate it with the name of the current file, so that
+  " we don't see the current file in our results.
+  let seen_set = {}
+  let current_file = fnamemodify(expand("%"), ":~:.")
+  let seen_set[current_file] = 1
+  " Start with the list of all open buffers, from most-recently-accessed to
+  " least-recently-accessed.
+  let paths = []
+  for bufnum in s:sorted_buffers()
+    let name = bufname(bufnum)
+    " Avoid calling fnamemodify on empty names. It produces a nonempty result.
+    if len(name) > 0
+      let path = fnamemodify(name, ":~:.")
+      if get(seen_set, path) == 0
+        call add(paths, path)
+        let seen_set[path] = 1
+      endif
+    endif
+  endfor
+  " Now append all the readable oldfiles, if we haven't seen them already.
+  for oldfile in filter(v:oldfiles, "filereadable(expand(v:val))")
+    " Do the same transformation we do to the buffer names.
+    let path = fnamemodify(oldfile, ":~:.")
+    " If the path is new, add it to both collections.
+    if get(seen_set, path) == 0
+      call add(paths, path)
+      let seen_set[path] = 1
+    endif
+  endfor
+  return paths
 endfunction
 
 function! s:history_source(type)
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -562,19 +590,25 @@ function! s:format_buffer(b)
...
+" from most-recently-accessed to least-recently-accessed
+function! s:compare_buffers(...)
   let [b1, b2] = map(copy(a:000), 'get(g:fzf#vim#buffers, v:val, v:val)')
   " Using minus between a float and a number in a sort function causes an error
-  return b1 > b2 ? 1 : -1
+  return b1 < b2 ? 1 : -1
+endfunction
+
+" from most-recently-accessed to least-recently-accessed
+function! s:sorted_buffers()
```
oconnor663: factoring out this function so that `s:history_files()` can use it.

junegunn:
Thanks. We already have `fzf#vim#_uniq` for deduplicating lists. So I believe this should suffice:

```diff
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index 42df1e0..cbb15b1 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -408,9 +408,11 @@ endfunction
 " History[:/]
 " ------------------------------------------------------------------
 function! s:all_files()
-  return extend(
-  \ filter(reverse(copy(v:oldfiles)), "filereadable(expand(v:val))"),
-  \ filter(map(s:buflisted(), 'bufname(v:val)'), '!empty(v:val)'))
+  return fzf#vim#_uniq(map(
+    \ [expand('%')]
+    \   + filter(map(s:buflisted_sorted(), 'bufname(v:val)'), '!empty(v:val)')
+    \   + filter(copy(v:oldfiles), "filereadable(expand(v:val))"),
+    \ 'fnamemodify(v:val, ":~:.")'))
 endfunction
 
 function! s:history_source(type)
@@ -464,8 +466,8 @@ endfunction
 
 function! fzf#vim#history(...)
   return s:fzf('history-files', {
-  \ 'source':  filter(reverse(s:all_files()), 'v:val != expand("%")'),
-  \ 'options': '-m --prompt "Hist> "'
+  \ 'source':  s:all_files(),
+  \ 'options': ['-m', '--header-lines', '1', '--prompt', 'Hist> ']
   \}, a:000)
 endfunction
 
@@ -565,16 +567,18 @@ endfunction
 function! s:sort_buffers(...)
   let [b1, b2] = map(copy(a:000), 'get(g:fzf#vim#buffers, v:val, v:val)')
   " Using minus between a float and a number in a sort function causes an error
-  return b1 > b2 ? 1 : -1
+  return b1 < b2 ? 1 : -1
 endfunction
 
-function! fzf#vim#buffers(...)
-  let bufs = map(sort(s:buflisted(), 's:sort_buffers'), 's:format_buffer(v:val)')
+function! s:buflisted_sorted()
+  return sort(s:buflisted(), 's:sort_buffers')
+endfunction
 
+function! fzf#vim#buffers(...)
   let [query, args] = (a:0 && type(a:1) == type('')) ?
         \ [a:1, a:000[1:]] : ['', a:000]
   return s:fzf('buffers', {
-  \ 'source':  reverse(bufs),
+  \ 'source':  map(s:buflisted_sorted(), 's:format_buffer(v:val)'),
   \ 'sink*':   s:function('s:bufopen'),
   \ 'options': '+m -x --tiebreak=index --header-lines=1 --ansi -d "\t" -n 2,1..2 --prompt="Buf> "'.s:q(query)
   \}, args)
```

Here I'm using `--header-lines 1` to make the current file unselectable. This happens to work nicely with the current buffer without a name which expands to the name of the working directory.

oconnor663:
Thanks for the quick reply! When I try your version in my home directory, with a new vim instance that hasn't opened anything yet, I find that the header line is actually a file I would've wanted to be selectable. However, if I move to a different directory, then the header line in that case is the cwd like you intended. I think the problem is that `fnamemodify(expand('%'), ":~:.")` from the blank buffer is an empty string if I'm not in a _subdirectory_ of my home folder. So it's blank for `/tmp` and `~`, but not blank for `~/foo`.

junegunn:
Okay, let's just unset `--header-lines` when `expand('%')` is empty. Let me know if you see any issues. Thanks!

oconnor663:
Thank you!

(Merged as e246016844f95a67b9602a2d3e6af24af8dbdbaf, original branch at https://github.com/oconnor663/fzf.vim/tree/mruhistory.)

PR #436, opened 2017-08-25
[fzf#vim#complete#path] works in Windows

janlazo:
Aside from `<plug>(fzf-complete-word)`, all of the other `<plug>`  are okay.
I used double quotes for `<plug>(fzf-complete-file-ag)` because single quotes for empty pattern doesn't work for ag in cmd.exe

Tested only in Vim and GVim.
For some reason, mappings don't work anymore in Neovim on Windows.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1171,7 +1171,7 @@ endfunction
 
 function! s:complete_trigger()
   let opts = copy(s:opts)
-  let opts.options = printf('+m -q %s %s', shellescape(s:query), get(opts, 'options', ''))
+  call s:merge_opts(opts, ['+m', '-q', s:query])
   let opts['sink*'] = s:function('s:complete_insert')
   let s:reducer = s:pluck(opts, 'reducer', s:function('s:first_line'))
   call fzf#run(opts)
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1255,9 +1255,10 @@ function! fzf#vim#complete(...)
     call s:merge_opts(s:opts, remove(s:opts, 'extra_options'))
   endif
   if has_key(s:opts, 'options')
-    " FIXME: fzf currently doesn't have --no-expect option
-    if type(s:opts.options) == s:TYPE.string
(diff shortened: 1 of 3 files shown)


autoload/fzf/vim.vim
```diff
@@ -1243,12 +1243,13 @@ function! fzf#vim#complete(...)
...
-    let s:opts.options =
-      \ join(filter([get(s:opts, 'options', ''), remove(s:opts, 'extra_options')], '!empty(v:val)'))
+    call s:merge_opts(s:opts, remove(s:opts, 'extra_options'))
   endif
   if has_key(s:opts, 'options')
     " FIXME: fzf currently doesn't have --no-expect option
-    let s:opts.options = substitute(s:opts.options, '--expect=[^ ]*', '', 'g')
+    if type(s:opts.options) == s:TYPE.string
+      let s:opts.options = substitute(s:opts.options, '--expect=[^ ]*', '', 'g')
+    endif
```
junegunn: fzf now supports `--no-expect` so we can simplify this part.

janlazo: Can you promote the fix on master so I have a reference on what you have in mind?
I'll rebase on `master` after.

junegunn: The line here is to make sure that there is no `--expect` option specified in the chain of options. Since fzf now has `--no-expect`, we can simply append `--no-expect` at the end of the string (of list).

```diff
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index c1ca760..f40baef 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1256,8 +1256,7 @@ function! fzf#vim#complete(...)
       \ join(filter([get(s:opts, 'options', ''), remove(s:opts, 'extra_options')], '!empty(v:val)'))
   endif
   if has_key(s:opts, 'options')
-    " FIXME: fzf currently doesn't have --no-expect option
-    let s:opts.options = substitute(s:opts.options, '--expect=[^ ]*', '', 'g')
+    let s:opts.options = .' --no-expect'
   endif
 
   call feedkeys("\<Plug>(-fzf-complete-trigger)")
```

junegunn:
Merged, thanks!

PR #441, opened 2017-09-03
[fzf#vim#preview] works with Windows default command

janlazo:
Ruby script doesn't work with cmd.exe in the terminal if the script is run in a batchfile.
I don't know how to preview the file in a batchfile or powershell so bash script is the only option. I adjusted the bash script to handle the network drive such that it works with `dir /s/b`.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -32,7 +32,7 @@ let s:is_win = has('win32') || has('win64')
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:bin_dir = expand('<sfile>:h:h:h').'/bin/'
 let s:bin = {
-\ 'preview': s:bin_dir.(executable('ruby') ? 'preview.rb' : 'preview.sh'),
+\ 'preview': s:bin_dir.(!s:is_win && executable('ruby') ? 'preview.rb' : 'preview.sh'),
 \ 'tags':    s:bin_dir.'tags.pl' }
 let s:TYPE = {'dict': type({}), 'funcref': type(function('call')), 'string': type(''), 'list': type([])}
 if s:is_win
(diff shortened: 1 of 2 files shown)


bin/preview.sh
```diff
@@ -15,6 +15,12 @@ fi
 
 IFS=':' read -r -a INPUT <<< "$1"
 FILE=${INPUT[0]}
+
+if [[ $1 =~ ^[A-Z]:\\ ]]; then
+  FILE=$FILE:${INPUT[1]}
+  shift
+fi
```
junegunn: Shouldn't we also change the value of `$CENTER`, to maybe `INPUT[2]`? And if I'm not mistaken, `shift` is not needed here.

janlazo: My intent is to shift `INPUT` to avoid additional changes and using `else` but I placed if statement in the wrong spot.

autoload/fzf/vim.vim
```diff
@@ -41,7 +41,7 @@ if s:is_win
   else
     let s:bin.preview = fnamemodify(s:bin.preview, ':8')
   endif
-  let s:bin.preview = escape(s:bin.preview, '\')
+  let s:bin.preview = 'bash '.escape(s:bin.preview, '\')
```
junegunn: What happens if we prepend `'ruby '` here when `executable('ruby')` is true? Can you elaborate more on why Ruby script cannot be used on Windows?

janlazo: Highlighting the current row didn't work when I ran the ruby script inside a batchfile.

My original plan is to make a batchfile, `preview.cmd`, to resolve the network drive via `cd /d` and then call the ruby script via `ruby .\preview.rb` or bash script via `bash .\preview.sh`. cmd.exe does not know about shebang and relies on associations between an extension and an executable to execute anything besides `*.exe`. I made one to resolve the network drive for both preview scripts without modifying them and they both run but highlighting didn't work for the ruby script.

What's strange is that highlight worked if the user associated the `rb` extension with ruby and then ran the preview script directly. It works for `fzf#vim#ag`.

junegunn: Okay, let me merge this for now. Please let me know if you happen to find a solution.

junegunn:
Thanks!

janlazo:
Reference for the batchfile wrapper for the preview scripts:
https://gist.github.com/janlazo/d906efbd61d61db7cae7eb010f70a439

PR #454, opened 2017-09-27
Remove hardcoded highlighting color

### LINKED ISSUE #452: Inconsistent theming in :Ag
- Category
    - [ ] Question
    - [x] Bug
    - [x] Suggestion
- OS
    - [x] Linux
    - [x] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [ ] Vim
    - [x] Neovim

Hi, it would be great if `:Ag` search respected the theme set by user. It can be done by removing '--color', 'hl:68,hl+:110' (line: 674 in autoload/fzf/vim.vim)

alfunx:
This option overrides the custom color settings in `FZF_DEFAULT_OPTS`, which should not happen. If that is not set by the user, it should use the default value of fzf (this seems to be the wrong location to set such a default).

Closes #452

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -683,8 +683,7 @@ function! fzf#vim#grep(grep_command, with_column, ...)
   \ 'source':  a:grep_command,
   \ 'column':  a:with_column,
   \ 'options': ['--ansi', '--prompt', capname.'> ',
-  \             '--multi', '--bind', 'alt-a:select-all,alt-d:deselect-all',
-  \             '--color', 'hl:68,hl+:110']
+  \             '--multi', '--bind', 'alt-a:select-all,alt-d:deselect-all']
   \}
   function! opts.sink(lines)
     return s:ag_handler(a:lines, self.column)


junegunn:
Yes, I totally agree with the change.

Those colors are set because on my terminal color scheme (seoul256), it's not possible to distinguish the color ag uses to display file name from the color fzf uses to highlight the matching substrings. It's my own specific problem, and it's true that it should be handled in a more general way, but then again, this is [my own plugin (or presets) designed to solve my own problems](https://github.com/junegunn/fzf.vim#rationale) :) And I don't want to make it go against my own preferences. Let me think more about this. Feel free to use your own fork for now.

alfunx:
That's fair enough. I'm using a consistent color scheme on my terminal and inside vim, that's why I thought this would be ideal. I also changed the file name highlighting color of Ag to make them less intrusive, maybe that would be an option for you.

By the way, thanks for all your work. FZF is awesome!

fenetikm:
I totally understand your reasoning on this, however, I would just like to say that this is also something that fixes up the styling / colouring to match when running Ag.

trevordmiller:
Related: https://github.com/junegunn/fzf.vim/issues/581

YUFENGWANG:
@junegunn what about Rg command? is this only for Ag?

tambourine-man:
> @junegunn what about Rg command? is this only for Ag?

I have the same question. Colors are bearable in Terminal.app but no in MacVim. 
Unreadable Magenta and Blue

GregSantulli:
@YUFENGWANG @tambourine-man I was able to get it working with rg: 

```vim
command! -bang -nargs=* Find call fzf#vim#grep('rg  --color always '.shellescape(<q-args>), 1, {'options': ['--color', 'hl:9,hl+:14']}, <bang>0)
```
Note the `with_column` argument. I have it set to `1` here, though I am not exactly sure what it is doing. https://github.com/junegunn/fzf.vim/blob/359a80e3a34aacbd5257713b6a88aa085337166f/autoload/fzf/vim.vim#L699

For those of you using `fzf#vim#with_preview()` options (as I am), these options can be passed into `with_preview()` like so:

```vim
command! -bang -nargs=* Find call fzf#vim#grep('rg  --color always '.shellescape(<q-args>), 1, fzf#vim#with_preview({'options': ['--color', 'hl:9,hl+:14']}), <bang>0)
```

PR #471, opened 2017-10-22
[fzf#vim#preview] Use ruby in Windows if available

### LINKED ISSUE #459: Enhancement: Use ruby preview script in Windows if ruby is available in PATH
- Category
    - [ ] Question
    - [ ] Bug
    - [X] Suggestion
- OS
    - [ ] Linux
    - [ ] macOS
    - [X] Windows
    - [ ] Etc.
- Vim
    - [X] Vim
    - [X] Neovim

See https://github.com/junegunn/fzf/issues/1018#issuecomment-333302497

cygwin bash requires double escaping, one for each shell, because fzf uses cmd.exe to run the preview command and the bash parses the passed arguments to the preview script.

Windows ruby does not reparse the arguments for another shell so only single escaping for cmd.exe is required. However, the ruby preview script cannot handle Windows-style absolute paths because of the network drive. Also, the ruby script requires recent fixes from the main repo for the ruby preview script to highlight correctly so this enhancment is for 0.17.1 (not released yet).

janlazo:
Close #459
Requires 0.17.1

This re-enables the ruby preview script if ruby is available in PATH. The changes is a port of the fix for Windows' network drive in the bash preview script.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -32,7 +32,7 @@ let s:is_win = has('win32') || has('win64')
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:bin_dir = expand('<sfile>:h:h:h').'/bin/'
 let s:bin = {
-\ 'preview': s:bin_dir.(!s:is_win && executable('ruby') ? 'preview.rb' : 'preview.sh'),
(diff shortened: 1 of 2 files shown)


junegunn:
Looks good to me, thanks!

PR #473, opened 2017-10-24
Avoid duplicate entries in :GFiles

joshpencheon:
This should crudely filter out files that are listed multiple times in `git ls-files` because of the state they're in (e.g. during a rebase).

Perhaps there is a better way to achieve this? I'm afraid I'm not very familiar with the internals of `fzf` / `fzf.vim`.

Thanks,
Josh

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -508,7 +508,7 @@ function! fzf#vim#gitfiles(args, ...)
   endif
   if a:args != '?'
     return s:fzf('gfiles', {
-    \ 'source':  'git ls-files '.a:args.' | uniq',
+    \ 'source':  'git ls-files '.a:args.(s:is_win ? '' : ' | uniq'),
     \ 'dir':     root,
     \ 'options': '-m --prompt "GitFiles> "'
     \}, a:000)
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -508,7 +508,7 @@ function! fzf#vim#gitfiles(args, ...)
   endif
   if a:args != '?'
     return s:fzf('gfiles', {
-    \ 'source':  'git ls-files '.a:args.' | sort | uniq',
+    \ 'source':  'git ls-files '.a:args.' | uniq',
     \ 'dir':     root,
     \ 'options': '-m --prompt "GitFiles> "'
     \}, a:000)
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -508,7 +508,7 @@ function! fzf#vim#gitfiles(args, ...)
   endif
   if a:args != '?'
     return s:fzf('gfiles', {
-    \ 'source':  'git ls-files '.a:args,
+    \ 'source':  'git ls-files '.a:args.' | sort | uniq',
     \ 'dir':     root,
     \ 'options': '-m --prompt "GitFiles> "'
     \}, a:000)


junegunn:
Thanks, I didn't know that the command can generate duplicate entries. Is there an easy way to reproduce the case?

> Perhaps there is a better way to achieve this?

I hope there's a better way because the pipeline will block at `sort` and will not produce any line until `git ls-files` is completed and `sort` has finished sorting the whole list. This can make the command unresponsive for a noticeable period of time if the repository has a large number of files. I'm not sure if it's worth it considering that this seems like a niche case.

joshpencheon:
Here's a contrived example:

```
mkdir fzf_vim_pr_473
cd fzf_vim_pr_473
git init .
echo 'a1' > a; echo 'b1' > b
git add a b; git commit -m "c1"
git checkout -b feature
echo 'a2' > a; echo 'b2' > b
git commit -am "c2"
git checkout master
echo 'a3' > a; echo 'b3' > b
git commit -am "c3"
git rebase feature    # trigger conflict
git ls-files          # see 3 versions of each 'a' and 'b', in the unmerged state
```

From `man git-ls-files`:
> For an unmerged path, instead of recording a single mode/SHA-1 pair, the index records up to three such pairs; one from tree O in stage 1, A in stage 2, and B in stage 3.

Although the window for this to occur is small (only during conflict), it is exactly a time when you'd be wanting to open the conflicting [duplicated] files...

`git ls-tree --name-only -r HEAD` doesn't show the duplicates, but then doesn't include staged-but-uncommited files.

Good point - I'd overlooked `sort` blocking. Anecdotally, the `ls-files` duplicates are contiguous, although I can't find documentation to confirm this is always the case. Maybe just passing through `uniq` as a best effort?

henrik:
I'm seeing a lot of duplicates with `:Files` (but not `:GFiles`) in a project:

screenshot 2017-10-30 12 14 24

Let me know if there's any command you'd like for me to run to help debug this. I tried `find . | grep item.rb` and it doesn't list duplicates, for example, but I haven't dug into the code yet to see what is going on.

Mentioning this in case it affects what solutions you will consider. Could be nice to have de-duplication in fzf independent of whether it's Files vs GFiles etc, if that could be done in a performant way.

joshpencheon:
@henrik, do you see the same duplicates using `fzf` from the command line? Are you setting a custom `FZF_DEFAULT_COMMAND` in your environment?

henrik:
@joshpencheon Yes, also see the duplicates with the `~/.fzf/bin/fzf` binary in the same project. Have not set a custom default command.

joshpencheon:
@henrik - in which case, sounds like it might be an issue worth raising in junegunn/fzf instead?

The "default-default" command that FZF runs is here:
https://github.com/junegunn/fzf/blob/master/src/constants.go#L58

That may help you diagnose where the duplicates are coming from.

junegunn:
@joshpencheon Thanks for the repro. `| uniq` seems like the most sensible solution for now, but I'm a little concerned how it'll fare on Windows. For the past months, @janlazo put a lot of effort into making the commands work on Windows, so I'd like to make sure that we don't break it.

janlazo:
cmd.exe doesn't have `uniq` equivalent but `uniq` can be emulated with powershell. Problem is powershell has a slow startup and always converts to utf16 strings on shell redirection. I encountered this issue while working on Windows support for fzf on nvim 0.2.0 and gave up on it. The user could upgrade powershell in Windows 7,8 but that's harder than installing git.

msysgit comes with `uniq` but cygwin utils may require double escaping if not running on bash.
Also,  `uniq` is in `/usr/bin` but git is in `/bin`. Cygwin users don't notice this if they use `git-bash.exe` because `/bin` and `/usr/bin` are treated as same folders on bash, not cmd.exe.

Unlike powershell, cmd.exe doesn't fully support utf8 so `uniq` may not work as expected for non-trivial strings.

cmd.exe has `sort` so the output can be passed to `for` loop but cmd.exe loops are synchronous so it does not scale to large codebases.

junegunn:
Thanks for the comment, that's quite complicated. Since I want to keep things simple, and having duplicate entries seems like a niche case, I suggest that we apply `uniq` only on non-windows platforms.

joshpencheon:
@junegunn I've added a conditional toggle, if that's ok?

junegunn:
LGTM, thanks.

Issue #389, opened 2017-06-11
:Tags, :BTags and gotags?


deanishe:
- Category
    - [x] Question
- OS
    - [x] Linux
    - [x] macOS
- Vim
    - [x] Neovim

`ctags` doesn't understand Go code. I'm trying to configure `fzf` so that `:Tags` and `:BTags` work for Go via `gotags`.

Does anyone know the correct commands?

I've had partial success with `let g:fzf_tags_command = 'gotags -f tags -R .'`, but only for `:Tags`, not `:BTags`.

junegunn:
Currently there's no option for changing the command for BTags. I could add a new configuration variable, or you could try implementing your own command from scratch using `fzf#run` function, but as [Universal ctags](https://github.com/universal-ctags/ctags) supports golang pretty nicely, I don't think it's necessary.

See http://docs.ctags.io/en/latest/news.html#new-parsers

deanishe:
Thanks for the great tip. I've installed universal ctags and it's working very well.

PR #477, opened 2017-10-26
Allow customizing btags command by file type

ViViDboarder:
This allows the optional setting of g:fzf_btags_command to a dict that
maps a filetype to a command string that will generate tags for the
current buffer.

Eg:
```
let g:fzf_btags_command = {'go': 'gotags -silent -sort %s'}
```
This is to resolve the issue seen in #389 in a similar way to CtrlP

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -747,9 +747,15 @@ function! fzf#vim#buffer_tags(query, ...)
   let args = copy(a:000)
   let escaped = fzf#shellescape(expand('%'))
   let null = s:is_win ? 'nul' : '/dev/null'
-  let tag_cmds = (len(args) > 1 && type(args[0]) != type({})) ? remove(args, 0) : [
-    \ printf('ctags -f - --sort=no --excmd=number --language-force=%s %s 2> %s', &filetype, escaped, null),
-    \ printf('ctags -f - --sort=no --excmd=number %s 2> %s', escaped, null)]
+  if (len(args) > 1 && type(args[0]) != type({}))
+      let tag_cmds = remove(args, 0)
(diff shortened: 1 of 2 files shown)


ViViDboarder:
The other thing I did to work around this was:
```
" Override BTags to attempt to include gotags as well
command! FZFBTags
            \ call fzf#vim#buffer_tags(<q-args>, [
            \ printf('gotags -sort -silent %s | sed /^!_TAG_/d', expand('%:S')),
            \ printf('ctags -f - --sort=no --excmd=number --language-force=%s %s 2> /dev/null', &filetype, expand('%:S')),
            \ printf('ctags -f - --sort=no --excmd=number %s 2> /dev/null', expand('%:S'))
            \ ], {})
```
The only things I don't like about this is that it tries gotags for everything and that I have to duplicate the other ctags commands that are already in the default function.

junegunn:
Thanks, but I'm currently not interested in adding any more options. You could try the following approach instead:

```vim
command! -bang -nargs=* BTags
  \  if &filetype == 'go'
  \|   call fzf#vim#buffer_tags(<q-args>, 'gotags -silent -sort '.shellescape(expand('%')), <bang>0)
  \| else
  \|   call fzf#vim#buffer_tags(<q-args>, <bang>0)
  \| endif
```

Or something like this, if you want to make it more generic:

```vim
command! -bang -nargs=* BTags
  \  if has_key(g:fzf_btags_command, &filetype)
  \|   call fzf#vim#buffer_tags(<q-args>, printf(g:fzf_btags_command[&filetype], shellescape(expand('%'))), <bang>0)
  \| else
  \|   call fzf#vim#buffer_tags(<q-args>, <bang>0)
  \| endif
```

ViViDboarder:
Thanks! I'll give one of those a go. At least this will show up when someone searches for how to use fzf with gotags.

PR #493, opened 2017-11-08
Escape invocation to preview.rb.

baz:
If the Vim runtime directory is located in a directory with spaces, preview will fail.


janlazo:
`fnameescape` should not be used in the shell and it won't work in Windows.

PR #500, opened 2017-11-12
[GitFiles] Support files with special characters

mbudde:
By default git will quote filenames that contain special characters.
From the git help on the `core.quotepath` config:

> The commands that output paths (e.g. `ls-files`, `diff`), when not given
> the `-z` option, will quote "unusual" characters in the pathname by
> enclosing the pathname in a double-quote pair and with backslashes the
> same way strings in C source code are quoted. If this variable is set
> to false, the bytes higher than 0x80 are not quoted but output as
> verbatim. Note that double quote, backslash and control characters are
> always quoted without `-z` regardless of the setting of this variable.

This quoting behaviour means that GitFiles cannot be used to open files
that contain any special characters. Fix this by using the `-z` option
to `ls-files` together with the `--read0` option on fzf.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -495,9 +495,9 @@ function! fzf#vim#gitfiles(args, ...)
   endif
   if a:args != '?'
     return s:fzf('gfiles', {
-    \ 'source':  'git ls-files '.a:args,
+    \ 'source':  'git ls-files -z '.a:args,
     \ 'dir':     root,
-    \ 'options': '-m --prompt "GitFiles> "'
+    \ 'options': '--read0 -m --prompt "GitFiles> "'
     \}, a:000)
   endif
 


junegunn:
Thanks, can you give me an example where it fails? I've never experienced the problem, so I'd like to see a real-world example.

mbudde:
Sure, I have run into the problem with filenames containing unicode:
```shell
$ git init fzftest && cd fzftest
$ touch testæøå.txt
$ git add .
$ git ci -m test
$ git ls-files | fzf --select-1
"test\303\246\303\270\303\245.txt"
$ git ls-files -z | fzf --read0 --select-1
testæøå.txt
```

junegunn:
Merged, thanks!

junegunn:
Unfortunately, this was not compatible with the recent change where we appended `uniq` to the command (#473) and I had to undo the change. I suggest that you define your own command using fzf#run and fzf#wrap.

```vim
command! -bang GF2 call fzf#run(fzf#wrap({'options': '--read0', 'source': 'git ls-files -z'}, <bang>0))
```

mbudde:
Aww, that sucks. I can see that my `uniq` has a `-z` flag but that looks like to be a GNU extension (it's not mentioned in [POSIX](https://www.unix.com/man-page/POSIX/1posix/uniq/) or [macOS](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/uniq.1.html) man pages). Would you accept a PR that adds a `--uniq` flag to FZF?

Another option could be passing the `-c core.quotepath=off` option to `git`. That gets rid of most quoting but not all as mentioned in the git help I posted.

I'll go with a custom command for now.

junegunn:
> Would you accept a PR that adds a --uniq flag to FZF?

No, it's not what I want in fzf. Anything that can be done outside of fzf is better be done that way.

PR #501, opened 2017-11-14
Add fzf#vim#rg_raw to enable fine grained ripgrep search

seb-mueller:
Although ripgrep is supported in fzf.vim, it is not possible (to my knowledge) to pass on custom parameters ( similar to issue #92 ). Most importantly I couldn't find a way to specify a path or a glob.

Analogously to fzf#vim#rg_raw, this would add the function fzf#vim#rg_raw

One could then define the following function to the .vimrc: 

    command! -bang -nargs=* Rgraw
      \ call fzf#vim#rg_raw(<q-args>,
      \   <bang>0 ? fzf#vim#with_preview('up:60%')
      \           : fzf#vim#with_preview('right:50%:hidden', '?'),
      \   <bang>0)

Ultimately, this would allow a "raw" ripgrep search by specifying a pattern and a path: 

    :Rgraw test ~/mypath

It also allows for more fine grained control exploiting the ripgrep parameters (note, "!" enables the preview window):

    Rgraw! -e pattern1 -e pattern -g *.txt ~/mypath

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -670,6 +670,14 @@ function! fzf#vim#ag_raw(command_suffix, ...)
   return call('fzf#vim#grep', extend(['ag --nogroup --column --color '.a:command_suffix, 1], a:000))
 endfunction
 
+" rg command suffix, [options]
+function! fzf#vim#rg_raw(command_suffix, ...)
+  if !executable('rg')
+    return s:warn('rg is not found')
+  endif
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but you can just use `fzf#vim#grep` as suggested in https://github.com/junegunn/fzf.vim#advanced-customization. Remove `shellescape` if you want to pass arguments as they are to the underlying command.

PR #517, opened 2017-11-23
Add Compilers and BCompilers command

midchildan:
This PR introduces new commands that enables you to go through the available presets for compiler-related settings.


junegunn:
Thanks, but I don't use those commands so I'm not sure if I want to have them. I'll leave this open for a while so that other users can see how it can be implemented.

- https://github.com/junegunn/fzf/blob/master/README-VIM.md
- https://github.com/junegunn/fzf/wiki/Examples-(vim)

juanibiapina:
Copying and pasting this example doesn't seem to work: https://github.com/junegunn/fzf/wiki/Examples-(vim)#search-lines-in-all-open-vim-buffers

I get only some gibberish.

PR #518, opened 2017-11-23
[B]Commit preview fails if using fish shell

nthapaliya:
First commit: prefer echo to pipe over redirect (fixes the bug mentioned in the title)

Second commit: [B]Commit preview will fail if the commit message has another long string of numbers such as a tracker story id, etc. This limits the output of grep to select the first match (which will be the commit hash, or nothing at all)


junegunn:
LGTM, thanks!

nthapaliya:
Thanks for the speedy response!

Issue #519, opened 2017-11-24
Vim :Tags command doesn't find strings with spaces


itsNikolay:
- Category
    - [ ] Question
    - [x] Bug
    - [ ] Suggestion
- OS
    - [ ] Linux
    - [x] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [x] Vim
    - [ ] Neovim

-------------------------------------------------------
Vim `:Tags` command doesn't find strings with spaces
For example I have a line in `tags` file:
```
# App Notifications	projects/app/notifications.md	/^# App Notifications$/;"	h
```
When I typing: 
- `# Notifications` - it doesn't find this line
- `# App` - it finds!
- `# App Notifications` - doesn't
- `Notifi` - doesn't

P.S.
Terminal `fzf` command finds all those cases
It looks like something in vim plugin

Can it be fixed with a local `.vimrc` fix ?
Thanks!

PR #522, opened 2017-11-25
[Tags] Configurable `fzf` command

itsNikolay:
Allows to override flags for `fzf` command to resolve issue when `--nth
1..2` flag is not enough (for example for tags with spaces)
See: https://github.com/junegunn/fzf.vim/issues/519

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -839,12 +839,24 @@ function! fzf#vim#tags(query, ...)
       break
     endif
   endfor
-  let opts = v2_limit < 0 ? '--algo=v1 ' : ''
+
+  let opts = []
+
+  if v2_limit < 0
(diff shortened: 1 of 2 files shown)


junegunn:
Commands can be redefined with arbitrary options using autoload functions.

```vim
command! -nargs=* -bang Tags call fzf#vim#tags(<q-args>, {'options': '--reverse'}, <bang>0)
```

itsNikolay:
@junegunn could you provide an example of `--nth 1..2` overriden with `--nth 1..5` ?
This one doesn't work:
```vim
command! -nargs=* -bang Tags call fzf#vim#tags(<q-args>, {'options': '--nth 1..5'}, <bang>0)
```

junegunn:
Works for me. You can even completely undo `--nth` with `--nth ..`.

PR #557, opened 2018-01-02
[fzf#vim#with_preview] workaround system() newline on Neovim

janlazo:
Related: https://github.com/neovim/neovim/issues/7788

On Vim, it's `ruby path\\to\\bin\\preview.rb`.
On Neovim, it's `ruby > path\\to\\bin\\preview.rb`

The patch reduces the output to a single line so the extra newline doesn't affect `s:bin.preview`.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -37,7 +37,7 @@ let s:bin = {
 let s:TYPE = {'dict': type({}), 'funcref': type(function('call')), 'string': type(''), 'list': type([])}
 if s:is_win
   if has('nvim')
-    let s:bin.preview = split(system('for %A in ("'.s:bin.preview.'") do echo %~sA'), "\n")[1]
+    let s:bin.preview = split(system('for %A in ("'.s:bin.preview.'") do @echo %~sA'), "\n")[0]
   else
     let s:bin.preview = fnamemodify(s:bin.preview, ':8')
   endif


junegunn:
Cryptic to me, but I'll just trust you on this :) Thanks!

janlazo:
See https://github.com/junegunn/fzf.vim/pull/429/files#r135373085

Prepending a statement with `@` prevents cmd.exe from printing it to stdout even with `echo on`.
See `@echo off` in https://github.com/junegunn/fzf/commit/669a6fee40a2d946ab35561c47e76d6a2216948c#diff-24d9d57e35df5afce7890f2400917919R51.
`echo off` tells cmd.exe to not output the statements after it and `@` prevents cmd.exe from echoing `echo off`.

As for Neovim bug, I don't know why it happens now but it didn't happen when I added Windows support for `fzf#vim#with_preview`.

janlazo:
@junegunn I assumed the wrong cause but the patch is fine. The `PROMPT` env var affects for loops in cmd.exe because, by default, it outputs each command to stdout, like a batchfile. I set a custom `PROMPT` in my vimrc if it's defined so it happens on Vim and Neovim if I run either from cmd.exe.
I typically run terminal Vim on powershell and run neovim through a batchfile.

A more robust patch is to do either:
1. Temporarily unset `$PROMPT` before running cmd.exe and then revert to previous value when done
2. Prepend each command with `@` so cmd.exe doesn't echo it.
3. Wrap the command like this: `@echo off & some_command & echo on`.

This PR did Option 2.

junegunn:
Thank you very much for the clarification.

Issue #553, opened 2017-12-22
Tags on Windows


zokeefe:
- Category
    - [x] Question
    - [ ] Bug
    - [ ] Suggestion
- OS
    - [ ] Linux
    - [ ] macOS
    - [x] Windows
    - [ ] Etc.
- Vim
    - [x] Vim
    - [ ] Neovim

First - wow! Can't believe how much faster this is then ctrlp. Even indexing ~500k files, then matching, is still faster! Great job!

Now, so I can remove my dependency on ctrlp, need to get this working with ctags! Can't seem to get it working though..

- Have fzf and fzf.vim installed - both working, for example, with :Files
- Windows 10, vim 8 with +terminal
- Have perl installed
- Typing :Tags opens %USERPROFILE%\vimfiles\bundle\fzf.vim\bin\tags.pl   (??!)
- Don't know enough Vimscript to know how to use fzf#vim#tags()
- Have a tags ctags file in the current directory

janlazo:
`perl.exe` is not associated with `*.pl` files on your system so Windows defaults to notepad.

janlazo:
For `Tags` source (https://github.com/junegunn/fzf.vim/commit/1e40de4f2d5655b76ca4b5b1e387e1034d1815fa#diff-7274b7c6c87526542ef9087f2e997dfaR811), explicitly running tags.pl on perl should fix the issue.

PR #559, opened 2018-01-02
[Tags] run perl for s:bin.tags

janlazo:
Close: https://github.com/junegunn/fzf.vim/issues/553

This should allow Windows users to not require associating `pl` file extension with perl and avoid opening notepad.exe (default behaviour).
Personally, I don't need it because I use the installer from http://strawberryperl.com/ but msysgit comes with perl.exe already.


junegunn:
Looks good to me, thanks!

PR #567, opened 2018-01-20
Fix Helptags command

okcompute:
If `tags` is set in the `wildignore` option, `:Helptags` command would
always return an empty list.

The proposed fix is to ignore `wildignore` on the `globpath` call.


junegunn:
Merged, thanks!

PR #590, opened 2018-03-01
Change :GGrep example to work at project level.

shea-parkes:
Do this so it matches the ~official :GFiles

You are of course welcome to decline this change.  I just found this much more useful.  I fooled around a little bit with actually moving this into the plugin proper, but my vimscript abilities weren't quite up to the job.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,9 @@ You can use autoload functions to define your own commands.
 " Command for git grep
 " - fzf#vim#grep(command, with_column, [options], [fullscreen])
 command! -bang -nargs=* GGrep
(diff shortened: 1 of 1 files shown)


junegunn:
I like the idea. By the way, have you tried setting `dir` option?

```vim
command! -bang -nargs=* GGrep
  \ call fzf#vim#grep(
  \   'git grep --line-number '.shellescape(<q-args>), 0,
  \   { 'dir': systemlist('git rev-parse --show-toplevel')[0] }, <bang>0)
```

If you're unfamiliar with it, see https://github.com/junegunn/fzf/blob/master/README-VIM.md#fzfrun

shea-parkes:
I was unfamiliar with that (very useful) option.  I tried it out and it works great.  I've gone ahead and added a commit to swap to that.  Thanks again for the great tool (and for being willing to respond to a noob).

junegunn:
Merged, thanks!

PR #593, opened 2018-03-07
Commands customization

mg979:
A way to customize commands by letting them fetch possible overrides by user'defined dictionaries, which contain the entries/arguments to replace


mg979:
Ok thanks, I didn't know it was already possible.

PR #616, opened 2018-04-10
[Tags][Lines][Buffers] use list type for options

janlazo:
`s:q()` is removed and `--query` is included directly in the options list.
Tested with latest version on Windows 8.1 and Ubuntu 16.04


junegunn:
Merged, thanks!

PR #620, opened 2018-04-14
buffer_tags: sort tags

justinmk:
This is nice for (ab)using :BTags to "browse" functions in a source file.

I assume `--sort=no` was used for performance, but on a single file maybe that doesn't matter?

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -745,8 +745,8 @@ function! fzf#vim#buffer_tags(query, ...)
   let escaped = fzf#shellescape(expand('%'))
   let null = s:is_win ? 'nul' : '/dev/null'
   let tag_cmds = (len(args) > 1 && type(args[0]) != type({})) ? remove(args, 0) : [
-    \ printf('ctags -f - --sort=no --excmd=number --language-force=%s %s 2> %s', &filetype, escaped, null),
-    \ printf('ctags -f - --sort=no --excmd=number %s 2> %s', escaped, null)]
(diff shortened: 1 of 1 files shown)


junegunn:
Performance was not a concern. The intention was to display tags in the order they appear in the file so that the user can see how the file is organized in a glance, using it like the minimap of sublime text.  But since I'm not an avid user of this command, I don't have a strong opinion on this. You find it more helpful when the list is sorted?

justinmk:
Personally I would use folds if I wanted the original order. Maybe others can comment, or revert if they complain :)

junegunn:
Let's see if anyone complains. Thanks.

junegunn:
Hmm, I just hope that the sorting was smarter. It would be nice if the tags are grouped by their enclosing classes, modules, or namespaces.

janlazo:
@junegunn Windows has its own `sort`.

junegunn:
@janlazo Oh, so I guess it doesn't work the same?

janlazo:
@junegunn Different flags/parameters. I'd restrict the change to use `sort` to Unix only, excluding Cygwin, or use `exepath()` to check if the absolute path contains `Windows\System32\sort`.

junegunn:
@janlazo Fixed it, thanks for the comment.

ayroblu:
Just in the middle of optimising my setup, wanted to use tagbar but it wasn't working (and not that ergonomic compared to fzf anyways) and saw this, I got my setup to look like this so far:
image

I'm going to try figure out removing sorting (seems easy enough from the PR, but a bit annoying if it's on the ctags level, and then grouping by types (as you've mentioned)

PR #626, opened 2018-04-27
fix #573

### LINKED ISSUE #573: Error on opening History
- Category
    - [x] Question
    - [ ] Bug
    - [ ] Suggestion
- OS
    - [x] Linux
    - [ ] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [ ] Vim
    - [x] Neovim

Have lastest fzf and fzf.vim versions and the problem stays with the minimal configuration.

Yestarday (after playing with vim fugitive but i'm not suggesting it's necessarily related) I started to have this error after `:History`:
```
Error detected while processing function <SNR>54_history[7]..fzf#vim#history[1]..<SNR>165_all_files:                                              
line    1:                                                                                                                                        
E16: Invalid range                                                                                                                                
E16: Invalid range
```
Then I press whatever key and I can use History normally. But it's annoying.

When I checked this on the minimal configuration it showed me a bit different error text:
```
Error detected while processing function <SNR>33_history[7]..fzf#vim#history[1]..<SNR>34_all_files:                                    
line    1:                          
E16: Invalid range                  
E16: Invalid range
```

Tried removing plugin folder and installing it again. Fruitless.

Have a guess that it might be related to these strange lines in History that appeared after playing with fugitive.
`term://.//10541:git add --patch -- my/path/to/file.rb`

haakenlid:
use `fnamemodify(v:val, ':p')` instead of `expand(v:val)`

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -433,7 +433,7 @@ function! s:all_files()
   return fzf#vim#_uniq(map(
     \ filter([expand('%')], 'len(v:val)')
     \   + filter(map(s:buflisted_sorted(), 'bufname(v:val)'), 'len(v:val)')
-    \   + filter(copy(v:oldfiles), "filereadable(expand(v:val))"),
+    \   + filter(copy(v:oldfiles), "filereadable(fnamemodify(v:val, ':p'))"),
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks!

PR #627, opened 2018-05-03
Escape delimiter character in :Commands

LexSong:
Unescaped character causes error on Windows.
Properly escaping the delimiter fixes this issue.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -866,7 +866,8 @@ endfunction
 " ------------------------------------------------------------------
 " Commands
 " ------------------------------------------------------------------
-let s:nbs = nr2char(0x2007)
+let s:nbscode = 0x2007
+let s:nbs = nr2char(s:nbscode)
 
 function! s:format_cmd(line)
   return substitute(a:line, '\C \([A-Z]\S*\) ',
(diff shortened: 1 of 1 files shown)


janlazo:
Isn't this a one-liner change?
Can you provide an example to reproduce the error with a minimal vimrc?

LexSong:
Minimal vimrc:

    call plug#begin('~/.vim/plugged')
    Plug 'junegunn/fzf'
    Plug 'junegunn/fzf.vim'
    call plug#end()
    
    " Error only happens when setting the encoding to utf-8
    set encoding=utf-8

OS: Windows 10
Default encoding: cp950
Vim version: 8.0.1806

janlazo:
If the delimiter is escaped with double quotes (aka `fzf#shellescape`), does it work? May need to use list instead of string for the fzf options.
If it's because of the newline, then https://github.com/junegunn/vim-plug/pull/755 could be related.
If it's because of `set encoding=utf-8`, then `readfile()` or `writefile()` is likely the problem.

janlazo:
If you're using vim-sensible, then it uses `set encoding=utf-8` on gvim. Is this happening on terminal Vim only?

LexSong:
`fzf#shellescape` doesn't work here:
https://drive.google.com/open?id=199tANoZKWwzcihc-CdYVQXbs6tCXCo_M

It happens both on terminal Vim and GVim.

It's a simple fix to avoid directly passing non-ascii character into Windows command line.

LexSong:
bat file generated by FZF which cause the error:
[VIFC34E.tmp.bat](https://github.com/junegunn/fzf.vim/files/1994078/VIFC34E.tmp.bat.txt)

It will work with CRLF newlines, or with an escaped delimiter.

Maybe the best way is fixing how FZF generates the batch files.

janlazo:
Lets go with the batchfile fix first before resorting to golang regex. `set encoding=utf-8` is required for multibyte environments regardless of the codepage else Vim becomes unusable.

Issue #303, opened 2017-02-03
Relative filepath completion


jonsmithers:
- Category
    - [X] Question
    - [ ] Bug
    - [X] Suggestion
- OS
    - [X] Linux
    - [X] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [X] Vim
    - [X] Neovim

I think it'd be awesome to have filepath completion relative to the current buffer. I've seen Deoplete [implement this ability](https://github.com/Shougo/deoplete.nvim/issues/109), and I use it constantly for relative html imports. But being able to fuzzy find a relative import would be much awesomer.

I'd like to either suggest this as a feature, or ask how I might make this work within my own vimrc.

I came _really close_ with the following vimrc code:

    inoremap <expr> <c-x><c-f> CompleteRelativePath()
    func! CompleteRelativePath()
        :cd %:h
        call fzf#vim#complete#path("find . -path '*/\.*' -prune -o -print \| sed '1d;s:^..::'")
        return ""
    endfu

It even works when you're autocompleting for a parent directory or sibling directory (i.e. `"../../bower_components/<C-X><C-F>`. The only problem is that I need to add `:cd -` to change back to the original working directory, but somehow that stops fzf from using the `%:h` as its working directory, even though the fzf call precedes the `:cd -` call.)

I hope I've effectively explained myself. Any ideas? Or thoughts?

junegunn:
How about doing `fzf#vim#complete#path(printf("cd %s && find ...", shellescape(expand('%:h'))))`?

jonsmithers:
oh....that's nifty. It works for completing within the buffer's current directory (`"./<c-x><c-f>`). 

However, it's somehow broken path completion for child and parent directories. Imagining the following project structure....

```
(project root)/
 ├── parent/
 │   ├── child/
 │   │   ├── c.txt
 │   │   └── d.tx
 │   └── CURRENT_BUFFER.txt
 ├── a.txt
 └── b.txt
```


Opening `CURRENT_BUFFER.txt` and typing `"./child/<c-x><c-f>"` or `"../<c-x><c-f>` will briefly flicker a panel that reads something like "_cd: ./child: no such file or directory_" or "_cd: parent: no such file or directory_".

FabioAntunes:
I was just searching for the same feature, but probably in a different way, I was thinking something like using the regular fuzzy file search, when we select a file, based on our current file create a relative path to that file.

[...]

PR #628, opened 2018-05-04
Add fzf-complete-path-relative

netoctone:
Hello @junegunn 

I've implemented a solution for a feature discussed in https://github.com/junegunn/fzf.vim/issues/303

Basic use case is to fuzzyfind a file and insert path to it that is relative to buffer.

It helps a lot with languages that do not have library paths (e.g. JavaScript).

Maybe it can help somebody save some time.

Cheers,

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,18 +216,19 @@ command! -bang -nargs=? -complete=dir Files
 Mappings
 --------
 
-| Mapping                            | Description                               |
-| ---                                | ---                                       |
-| `<plug>(fzf-maps-n)`               | Normal mode mappings                      |
-| `<plug>(fzf-maps-i)`               | Insert mode mappings                      |
-| `<plug>(fzf-maps-x)`               | Visual mode mappings                      |
-| `<plug>(fzf-maps-o)`               | Operator-pending mappings                 |
-| `<plug>(fzf-complete-word)`        | `cat /usr/share/dict/words`               |
-| `<plug>(fzf-complete-path)`        | Path completion using `find` (file + dir) |
-| `<plug>(fzf-complete-file)`        | File completion using `find`              |
-| `<plug>(fzf-complete-file-ag)`     | File completion using `ag`                |
-| `<plug>(fzf-complete-line)`        | Line completion (all open buffers)        |
-| `<plug>(fzf-complete-buffer-line)` | Line completion (current buffer only)     |
+| Mapping                              | Description                                 |
+| ---                                  | ---                                         |
+| `<plug>(fzf-maps-n)`                 | Normal mode mappings                        |
+| `<plug>(fzf-maps-i)`                 | Insert mode mappings                        |
+| `<plug>(fzf-maps-x)`                 | Visual mode mappings                        |
+| `<plug>(fzf-maps-o)`                 | Operator-pending mappings                   |
+| `<plug>(fzf-complete-word)`          | `cat /usr/share/dict/words`                 |
+| `<plug>(fzf-complete-path)`          | Path completion using `find` (file + dir)   |
+| `<plug>(fzf-complete-path-relative)` | Generate relative path (starting ./ or ../) |
+| `<plug>(fzf-complete-file)`          | File completion using `find`                |
+| `<plug>(fzf-complete-file-ag)`       | File completion using `ag`                  |
+| `<plug>(fzf-complete-line)`          | Line completion (all open buffers)          |
+| `<plug>(fzf-complete-buffer-line)`   | Line completion (current buffer only)       |
 
 ### Usage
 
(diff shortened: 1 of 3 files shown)


rod-stuchi:
High expectation about this PR

kibs:
This works really nice!

dementor5:
I am also looking forward to the request. This function is needed!

oyren:
+1

chmanie:
I made an up-to-date fork using this PR and changed the command to `[fd](https://github.com/sharkdp/fd)` like so (to include directories):

```
inoremap <expr> <c-x><c-f> fzf#vim#complete#path_relative('fd')
```

And I can say it works really, really well. Please add this in to upstream, that'd make a really useful addition.

ybbond:
> I made an up-to-date fork using this PR and changed the command to [`fd`](https://github.com/sharkdp/fd) like so:
> 
> ```
> inoremap <expr> <c-x><c-f> fzf#vim#complete#path_relative('fd')
> ```
> 
> And I can say it works really, really well. Please add this in to upstream, that'd make a really useful addition. As far as I can see so far, there's no vim plugins or scripts that can adequately complete relative paths. The combination of fzf, fd and this PR is the best experience of completing paths I've ever had in an editor.

is there anyway I can implement this by adding your PR's script to my `init.vim`?
I have tried, but my experience in vim script is beyond terrible, I failed

chmanie:
@ybbond sure, there's not much to it. Here are the three lines in my `.vimrc`:

```vim
" I use plug
Plug '/usr/local/opt/fzf'
Plug 'chmanie/fzf.vim'
inoremap <expr> <c-x><c-f> fzf#vim#complete#path_relative('fd')
```

ybbond:
@chmanie yeah, I meant using `junegunn`'s fzf. Good idea tho, `fzf-addons-relative-path.vim` sounds neat.
but considering your repo is just 1 commit behind `junegunn`, I am using yours for now!

btw, I noticed your location. Enjoy Indonesia!

DavisOwen:
@junegunn are there any plans to merge this?

pvonmoradi:
@chmanie This hasn't merged yet, right?

dragonofdeath:
Here is a script if you want to use it with current fzf.vim plugin ( I adopted it to js/ts, but the idea stays the same) https://github.com/dragonofdeath/dotfiles/blob/master/nvim/.vim/scripts/fzf_js_import.vim

cideM:
I think this also does the trick, no?

```
  inoremap <expr> <c-x><c-f> fzf#vim#complete("fd <Bar> xargs realpath --relative-to " . expand("%"))
```

zorgick:
@cideM could you please explain, what this commands does? It works, that's great! It bothers me that, I was trying to solve the same problem via `fd` options, but couldn't find any mention about relative path search. 
I'm curious about this fragment `xargs realpath --relative-to` - what is it?

rod-stuchi:
@cideM this trick was amazing, well done Sir 👏👏.

@junegunn maybe worth it a mention on the [wiki vim examples](https://github.com/junegunn/fzf/wiki/Examples-(vim)) ?

ejhusom:
Here's a slight adjustment to the excellent command of @cideM, in order to deal with filenames that have spaces:

```vim
inoremap <expr> <c-x><c-f> fzf#vim#complete("find -print0 <Bar> xargs --null realpath --relative-to " . expand("%:h"))
```
In the line above I use `find` instead of `fd`, and have added `-print0` to make sure that `find` returns a null character at the end of each filename. By passing `--null` to `xargs`, we ensure that `xargs` know that each input item is terminated by a null character.

Here's a version using `rg` instead (which uses the flag `--null` instead of `-print0`):

```vim
inoremap <expr> <c-x><c-f> fzf#vim#complete("rg --files --hidden --no-ignore --null <Bar> xargs --null realpath --relative-to " . expand("%:h"))
```

To answer the question of @zorgick: `xargs realpath --relative-to` takes the output of `fd` (or `find` or `rg` etc), and uses `realpath`, a GNU coreutil, to produce the path of each match relative to the current file you have open in Vim. The `expand("%:h")` tells Vim to output the filepath of the current file, relative to the directory that you opened it from.

awelormro:
Maybe it's an old thread but I've made this function to insert the filename in the current buffer position

'''vim
function! InsertFileName()
  let selected_file = fzf#run({'source': 'find . -type f'})
  if !empty(selected_file)
    let selected_file_str = join(selected_file, "\n")
    execute "normal i" . selected_file_str
  endif
endfunction
'''

thoughtsunificator:
Tested the PR and it fails when you try to complete an existing path that begins with "./" for example, instead of completing the path relatively to the opened file, it resolves the path and you get a path that is relative to the project. 

```
import { Foo } from "./f" // press c-x c-f with cursor positionned after f letter.
```

In contrast, the following works as described:

```
import { Foo } from "f" // press c-x c-f with cursor positionned after f letter.
```

PR #637, opened 2018-05-22
Clean up buffer names in :Buffers

alaaibrahim:
Issue: While following paths in a file, it might end up in paths like
`a/b/../c/d.ext` or `a/b/./c.ext`, which would be cleaner if displayed
like `a/c/d.ext` and `a/b/c.ext`.

Solution: Expand fullpath to resolve directories before collapsing it to
home directory/relative path.


junegunn:
Merged, thanks!

PR #652, opened 2018-06-06
[Buffers] Add a custom filename_modifier for "Buffers" command

adamtaylor13:
This allows users to specify in their .vimrc a specific
filename_modifier in order to customize how their :Buffers list
looks. I didn't need the full path, so I wanted to be able to view
just the filenames themselves.

This allows users to specify something like this in their .vimrc:
`let g:fzf_buffer_fnamemodify_arg=":t"`

This would make the `:Buffers` command show up like:
`> [1] # foo.txt`
Instead of the standard:
`> [1] # /path/to/foo.txt`

I also couldn't find any contribution guidelines. I'd be happy to format anything as required.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -576,8 +576,9 @@ function! s:bufopen(lines)
 endfunction
 
 function! s:format_buffer(b)
+  let filename_modifier = get(g:, 'fzf_buffer_fnamemodify_arg', ":~:.")
   let name = bufname(a:b)
-  let name = empty(name) ? '[No Name]' : fnamemodify(name, ":p:~:.")
+  let name = empty(name) ? '[No Name]' : fnamemodify(name, filename_modifier)
   let flag = a:b == bufnr('')  ? s:blue('%', 'Conditional') :
           \ (a:b == bufnr('#') ? s:magenta('#', 'Special') : ' ')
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -576,8 +576,9 @@ function! s:bufopen(lines)
 endfunction
 
 function! s:format_buffer(b)
+  let filename_modifier = get(g:, 'fzf_buffer_fnamemodify_arg', ":~:.")
```
adamtaylor13: You can set the `g:fzf_buffer_fnamemodifier_arg` variable in your .vimrc. Otherwise, it will default to the existing filename_modifier: ":~:."

adamtaylor13:
Is this repository still active? I'd like to contribute if at all possible. This was well over 2 months ago.

junegunn:
Yes, this repository is still active, but as [hinted on the README page](https://github.com/junegunn/fzf.vim#rationale), it is basically a collection of personal scripts, and I decided not to add options and features that I don't actually need. I'm sorry about that, but I just don't have time and mental bandwidth to maintain the code that I don't use, so please understand.

bridgesense:
Same here.  Hate to fork this project for such a small bit of code, but this is a serious QOL update.

bridgesense:
Just an FYI, this solution appears to break the preview since the filename view appears also to be the resource.

PR #693, opened 2018-09-06
Unfold after Lines/BLines

dsh2:
Actually I would want the unfold to only happen, if the matched line is really hidden within a fold. In the current version of the PR, the fold will also be unfolded when the matched line is the beginning of the fold. Unfortunately, I did not find any way to tell, if the current line is hidden within a fold.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -316,7 +316,7 @@ function! s:line_handler(lines)
   let keys = split(a:lines[1], '\t')
   execute 'buffer' keys[0]
   execute keys[2]
-  normal! ^zzzv
+  normal! ^zvzz
 endfunction
 
 function! fzf#vim#_lines(all)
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I think it would be better to do `zz` after `zv`, so `zvzz`. What do you think?

dsh2:
Yes! That's more natural - I think. I actually tend to have scrolloff=999 so I did not notice. By the way: Maybe you have an idea how one could check for the above mentioned problem with the first line?

dsh2:
I fear I don't get it. Maybe you mean `==` instead of `&&`? 

Even worse, I don't get what I meant in my first message of the PR. I think I mixed up the first line of a fold with the result of `foldtext()` -- which often times is very similar to the first line of a fold. 

Indeed `zv` just opens enough folds to make the line visible, i.e.not folded, in which the cursor is located at. So we could simply go for the `zvzz`-approach.

junegunn:
Okay, let's keep it simple.

PR #694, opened 2018-09-07
function to fill quickfix, some changes to BLines

mg979:
Commits:

1. common function to fill quickfix, since it's called in several places in the file
2. BLines accepts `--multi` and can build quickfix
3. BLines query isn't inserted in the fzf window, but it's used to prefilter lines

Quickfix for BLines allows `:cdo`.

With commit 3. you can do things like:
```
BLines {{{[0-9]
```
to filter foldmarkers.

In general, having `query` work as a prefilter allows to filter fzf list with some other text (much more freely that is), and it's also treated as a regex.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -388,6 +388,12 @@ function! s:buffer_line_handler(lines)
   if len(a:lines) < 2
     return
   endif
+  let qfl = []
+  for line in a:lines[1:]
+    let [ln, ltxt] = split(line, "\t")
+    call add(qfl, {'filename': expand('%'), 'lnum': str2nr(ln), 'text': ltxt})
+  endfor
+  call s:fill_quickfix(qfl, 'cfirst')
   normal! m'
   let cmd = s:action_for(a:lines[0])
   if !empty(cmd)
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -396,9 +413,9 @@ function! fzf#vim#buffer_lines(...)
   let [query, args] = (a:0 && type(a:1) == type('')) ?
         \ [a:1, a:000[1:]] : ['', a:000]
   return s:fzf('blines', {
-  \ 'source':  s:buffer_lines(),
+  \ 'source':  empty(query) ? s:buffer_lines() : filter(s:buffer_lines(), 'v:val =~ "'.query.'"'),
```
junegunn: Filter function here should go inside `s:buffer_lines`, so that it is applied on unformatted, original lines.

autoload/fzf/vim.vim
```diff
@@ -258,6 +258,17 @@ function! s:warn(message)
   return 0
 endfunction
 
+function! s:fill_quickfix(list, ...)
+  if len(a:list) > 1
+    call setqflist(a:list)
+    copen
+    wincmd p
+    if a:0
+      exe a:1
```
junegunn: A minor nit. Please change `exe` to `execute` for consistency.

junegunn:
Merged, thanks!

mg979:
Thanks :)

mg979:
Unfortunately the last change doesn't work... Because line numbers don't correspond anymore.

I had to change the function to make it work, but it's a bit more verbose:
```vim
function! s:buffer_lines(query)
  let linefmt = s:yellow(" %4d ", "LineNr")."\t%s"
  if empty(a:query)
    return map(getline(1, "$"), 'printf(linefmt, v:key + 1, v:val)')
  endif
  let lines = getline(1, "$")
  let ok = []
  let i = 1
  for l in lines
    if match(l, a:query) >= 0
      call add(ok, printf(linefmt, i, l))
    endif
    let i += 1
  endfor
  return ok
endfunction
```

What should be done?

junegunn:
Fixed it, thanks.

mg979:
Thanks again :)

PR #704, opened 2018-09-22
Ignore filename for grep

andreypopp:
There is no reason to search through filenames, otherwise you get too many false
positives.


TrySound:
@junegunn We are currently stuck with Andrey's fork. Would be good to use original project. Is this good for you?

junegunn:
I always search against file names, especially file extensions, so I won't merge this. And as documented on the README page, you can pass any options to the function, so merging this is not necessary.

```vim
command! -nargs=* Rg
  \ call fzf#vim#grep(
  \   'rg --column --line-number --no-heading --color=always --smart-case '.shellescape(<q-args>),
  \   1,
  \   {'options': '--delimiter : --nth 2..'})
```

TrySound:
@junegunn But why `Files` then? I use Files to search through file names and Ag to search through content. Why mix them in one?

PR #707, opened 2018-09-29
Configurable preview highlighter for Linux

gbcreation:
I would like to pass custom options to highlighters used for previews (for example to select a specific color theme for `highlight`).

This PR is to add a new configuration variable `g:fzf_preview_highlighter` that allows to specify which highlighter to use with specific command line options.

**The implementation works only on Linux (successfully tested) and probably on Mac (can someone test and confirm please?).** I don't have Windows, would be glad if someone could adapt changes for it :)

Example:

```vim
let g:fzf_preview_highlighter = "highlight -O xterm256 --line-number --style rdark --force"
```
allows to use `highlight` with the `rdark` theme.

If the `g:fzf_preview_highlighter` variable is not set,  the default behavior is similar to `preview.rb`: check if `highlight`, `coderay`, `rougify` and `cat` exist, and use the first one available.

There is also a new `g:fzf_preview_line_highlight` variable that allows to specify the ansi color code used to highlight the line in the preview window that corresponds to the selected one in the list. Default value is `\x1b[7m` ("REVERSE", the same color code used in `preview.rb`). 

Examples:

To highlight the line with the bright black background color, use:
```vim
let g:fzf_preview_line_highlight = '\x1b[101m'
```

For terminals supporting 256-colors, you can specify a 8-bit color code:
```vim
let g:fzf_preview_line_highlight = '\x1b[48;5;240m'
```

You can even specify a 24-bit (true color) color code for terminals supporting it like Konsole:
```vim
let g:fzf_preview_line_highlight = '\x1b[48;2;80;80;80m'
```

Screenshot:

vim_enhanced_fzf_preview

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -32,16 +32,32 @@ let s:is_win = has('win32') || has('win64')
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:bin_dir = expand('<sfile>:h:h:h').'/bin/'
 let s:bin = {
-\ 'preview': s:bin_dir.(executable('ruby') ? 'preview.rb' : 'preview.sh'),
 \ 'tags':    s:bin_dir.'tags.pl' }
 let s:TYPE = {'dict': type({}), 'funcref': type(function('call')), 'string': type(''), 'list': type([])}
 if s:is_win
+  let s:bin.preview = s:bin_dir.(executable('ruby') ? 'preview.rb' : 'preview.sh'),
   if has('nvim')
     let s:bin.preview = split(system('for %A in ("'.s:bin.preview.'") do @echo %~sA'), "\n")[0]
   else
     let s:bin.preview = fnamemodify(s:bin.preview, ':8')
   endif
-  let s:bin.preview = (executable('ruby') ? 'ruby' : 'bash').' '.escape(s:bin.preview, '\')
+  let s:bin.preview = (executable('ruby') ? 'ruby' : 'bash').' '.escape(s:bin.preview, '\').' {}'
+else
+  if exists("g:fzf_preview_highlighter")
+    let s:preview_highlighter = g:fzf_preview_highlighter
+  elseif executable('highlight')
+    let s:preview_highlighter = 'highlight -O ansi --line-numbers --force'
+  else
+    for s:hl in ['coderay', 'rougify', 'cat']
+      if executable(s:hl)
+        let s:preview_highlighter = s:hl
+        break
+      endif
+    endfor
+  endif
+  echom s:preview_highlighter
+  let s:preview_line_highlight = get(g:, 'fzf_preview_line_highlight', '\x1b[7m')
+  let s:bin.preview = 'CENTER={2};FIRST=$(($CENTER-$LINES/4));FIRST=$(($FIRST<1?1:$FIRST));'.s:preview_highlighter.' {1}|sed -n "$FIRST,+50{${CENTER}{s/\x1b\[[0-9;]*m/&' . s:preview_line_highlight . '/g;s/$/\x1b[m/};p}"'
 endif
 
 let s:wide = 120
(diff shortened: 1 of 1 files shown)


wookayin:
This is a nice feature. However I have a small comments and concern with its implementation:

* Most changes are on fzf.vim sources rather than on `preview.rb` or `preview.sh`, and you have copied the highlighter logic from those scripts into vim plugin scripts. In this case, we should avoid a redundant implementation.
* I very like having a configuration point such as `g:fzf_preview_highlighter`, which can be used to specify any program that has a similar behavior as `preview.rb` (e.g. `highlighter[:LINENO][:IGNORED]). Given that, I believe the reset of lines L48~L59 are not necessary. As in previous, we can simply keep L35 (where `s:bin.preview` is set).
* Having another configuration point `g:fzf_preview_line_highlight` is awesome. I'd suggest to pass this variable in terms of command-line flag or environment variable so that `preview.rb` or `preview.sh` can be aware of this escape sequence.

Great idea!

gbcreation:
Hi @wookayin! Thanks a lot for your feedback.

Glad you appreciate this feature.

Actually, my changes could be considered as a factorization of the logic you're talking about :slightly_smiling_face:
Let me explain that: currently, this logic is already duplicated into both `preview.rb` and `preview.sh` files. Every time you want to modify or enhance this logic, you have to make changes to both files and test both files. With my changes, all this logic is moved to a unique place in `fzf.vim`. It's easier to modify and test.

In fact, if you work on Linux (and probably on MacOs), you could remove the `preview.rb` and `preview.sh` files as they are no longer needed with my changes. I keep them only for compatibility with Windows platforms (that's why any help would be welcome to adapt my changes for Windows :wink: )

junegunn:
Thanks for your interest in the project. I'm not really into the idea of introducing and maintaining more configuration knobs for `fzf#vim#with_preview()` function, which can give the impression to the users that they have to use the function and tinker with those variables for setting up previewer when it's a purely optional helper function whose only role is to add a few more options to the dictionary passed as the argument.

```vim
echo fzf#vim#with_preview({})

" See https://github.com/junegunn/fzf/blob/master/README-VIM.md#fzfwrap
let opts = { 'source': 'ls', 'window': 'tabnew' }
echo fzf#wrap(opts)
echo fzf#vim#with_preview(fzf#wrap(opts))
```

It's quite trivial to replicate what it does without using it if you know about the options. For most users, `--preview` option is all they need.

```vim
" Files with bat previewer
command! -bang -nargs=? -complete=dir Files
  \ call fzf#vim#files(<q-args>, {'options': ['--preview', 'bat -p --color always {}']}, <bang>0)

" Rg using custom preview script
command! -bang -nargs=* Rg
\ call fzf#vim#grep(
\   'rg --column --line-number --no-heading --color=always --smart-case '.shellescape(<q-args>), 1,
\   'options': ['--preview', '~/bin/my-custom-preview-script {}']
\   <bang>0)
```

> this logic is already duplicated into both preview.rb and preview.sh

True, we don't need both scripts. `sh` version was added later by a contributer to support systems without Ruby. Currently it doesn't use an external syntax highlighters unlike the Ruby version, but I believe we can extend it and cover all use cases and eventually remove the original Ruby script.

And that's what I think we should do. I still prefer to have an external script file instead of having some shell script embedded somewhere in the bowels of a VimScript file. That way it's easier to test and maintain given that most users are not familiar with VimScript. You can use it outside of Vim separately, and easily copy and modify it to your liking.

wookayin:
+1 for still having an external script for previewing.

I suggest we should be able to simply configure which highlighter (e.g. bat, coderay, etc.) should be used (as proposed with `g:fzf_preview_highlighter`, etc.).

Using `--preview` (with custom preview scripts) in `fzf#vim#...()` is so flexible enough, but it still can be cumbersome to have a redundant implementation to support the `LINENO` feature (e.g. `preview.rb:LINENO:IGNORED`).

PR #712, opened 2018-10-07
Support bat for syntax highlighting in preview

wookayin:
[bat](https://github.com/sharkdp/bat) is a very nice clone of `cat` that supports syntax highlighting feature. If bat is available, one may want to have it as a syntax highlighter used in `preview.sh` or `preview.rb`. This PR adds the support.

I have implemented the functionality to be close to what `preview.sh` is doing, with `awk`. Of course, the use of `bat` is absolutely optional. It allows us to have a syntax highlighting feature even without ruby.

image

Another note: it might conflict with #707.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,10 +185,10 @@ command! -bang Colors
 " Augmenting Ag command using fzf#vim#with_preview function
 "   * fzf#vim#with_preview([[options], preview window, [toggle keys...]])
 "     * For syntax-highlighting, Ruby and any of the following tools are required:
+"       - Bat: https://github.com/sharkdp/bat
 "       - Highlight: http://www.andre-simon.de/doku/highlight/en/highlight.php
 "       - CodeRay: http://coderay.rubychan.de/
 "       - Rouge: https://github.com/jneen/rouge
-"       - Bat: https://github.com/sharkdp/bat
 "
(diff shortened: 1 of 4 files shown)


junegunn:
Thanks. Now with a few more changes, we no longer need the ruby version, right?

wookayin:
I agree with you that we can eventually remove the ruby script. One may still want to use `highlight`, `coderay`, or `rougify`, but as long as we are able to configure which highlighter is used (as well as a few more changes), the ruby one is not needed and therefore the previewer scripts can be unified into more general one.

For configuration points, let's discuss on https://github.com/junegunn/fzf.vim/pull/707#issuecomment-432494703.

junegunn:
Trying out bat, highlight, ..., finally cat in turn should be good enough for most users. It works out of the box without any configuration, and that's a big plus from my perspective. It's a little hard for me to imagine that someone installed `bat` or `highlight` would still want to use slower – or less "modern" – alternatives like `coderay` or `rougify`. `bat` even checks `BAT_THEME` environment variable, so one can still customize its look even though we don't allow custom commands.

felipesere:
This looks brilliant, what is missing for this to get merged? 😄

junegunn:
Made a few changes. Please see the commit message for details. Thanks.

wookayin:
Looks very good, thanks!

PR #737, opened 2018-11-14
Use `env` to get `bash`

tshakah:
This wasn't working on NixOS due to bash not being under `/bin`. Using `env` should make it compatible with more setups.


junegunn:
Merged, thanks.

PR #738, opened 2018-11-17
add g:fzf_lines_jump option to jump to existing viewport for Lines

pierreN:
add a new option like fzf_buffers_jump to switch to an existing window if present when using the Lines command


fuzzbomb:
This could be more powerful if `s:find_open_window` respected the `switchbuf` setting (or took arguments similar to `switchbuf`).

Maybe that warrants a seperate issue?

pierreN:
If you can get a PR merged which does that I'd be happy to close this one :+1:

PR #739, opened 2018-11-17
Fix preview of symbolic links

mimadrid:
First al all, thank so much @junegunn for your awesome contributions, keep it up!

I would like to see the symbolic links in the preview

I hope to be useful


junegunn:
Merged, thanks!

PR #745, opened 2018-12-06
Use `{count}tabnext` instead of `normal {count}gt`

mars90226:
Use `{count}tabnext` instead of `normal {count}gt` to avoid error that cannot re-enter normal mode in terminal mode in neovim.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -583,7 +583,7 @@ function! s:find_open_window(b)
 endfunction
 
 function! s:jump(t, w)
-  execute 'normal!' a:t.'gt'
+  execute a:t.'tabnext'
   execute a:w.'wincmd w'
 endfunction
 


junegunn:
Can you provide more details on the problem you're trying to fix here? Is it a bug of Neovim?

mars90226:
When I use `:Windows` command in a terminal buffer and select an item, neovim will say _"Vim(normal):Can't re-enter normal mode from terminal mode"_.

My neovim version is NVIM v0.3.2-873-g05f9c7c2f
Here is the minimal vimrc:
test_vimrc
And this is the error I get:
neovim_terminal_fzf_vim_windows_error

I've got the idea from [neovim  #4895](https://github.com/neovim/neovim/issues/4895) that `normal! gt` might be the problem that can be replaced by `tabnext`.

junegunn:
Thanks for the explanation, merged.

PR #768, opened 2019-02-20
Do not remap keys when using feedkeys

gpanders:
I have `;` and `:` swapped in my vimrc, so using the `:Commands` command is broken because `feedkeys` uses my remapped `:`. This PR tells `feedkeys` to not remap keys for the `:Commands` sink function.


junegunn:
Merged, thanks.

PR #774, opened 2019-02-26
expand '~' in preview

zhou13:
In preview mode, with this pull request it can correctly handle `~` in the path. Using `:call fzf#vim#history(fzf#vim#with_preview({'options': '--no-sort'}))` to test.

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -17,6 +17,7 @@ if [[ $1 =~ ^[A-Z]:\\ ]]; then
   CENTER=${INPUT[2]}
 fi
 
+FILE="${FILE/#\~/$HOME}"
 if [ ! -r "$FILE" ]; then
   echo "File not found ${FILE}"
   exit 1


zhou13:
Will the pull requests be merged?

junegunn:
Please make sure that the code only replaces `~/`, it won't work with `~username`.

zhou13:
To do that correctly, I think the correct way is to use 
````
expandPath() {
  local path
  local -a pathElements resultPathElements
  IFS=':' read -r -a pathElements <<<"$1"
  : "${pathElements[@]}"
  for path in "${pathElements[@]}"; do
    : "$path"
    case $path in
      "~+"/*)
        path=$PWD/${path#"~+/"}
        ;;
      "~-"/*)
        path=$OLDPWD/${path#"~-/"}
        ;;
      "~"/*)
        path=$HOME/${path#"~/"}
        ;;
      "~"*)
        username=${path%%/*}
        username=${username#"~"}
        IFS=: read _ _ _ _ _ homedir _ < <(getent passwd "$username")
        if [[ $path = */* ]]; then
          path=${homedir}/${path#*/}
        else
          path=$homedir
        fi
        ;;
    esac
    resultPathElements+=( "$path" )
  done
  local result
  printf -v result '%s:' "${resultPathElements[@]}"
  printf '%s\n' "${result%:}"
}
````
from https://stackoverflow.com/questions/3963716/how-to-manually-expand-a-special-variable-ex-tilde-in-bash.  Any suggestions?

junegunn:
No, full-fledged support for tilde expansion is too much. I'm just asking you to update the code to replace `~/` to `$HOME/`.

junegunn:
Fixed in #892.

PR #784, opened 2019-03-20
Fix buffer list update when using same window.

noboruma:
Change the buffer list update  trigger from WinBufEnter to BufEnter.
Without this change, switching between two buffers, from A to B, on the same window is broken. The buffer list is not updated once the buffer B is selected. Thus, the non-selectable buffer is still pointing at the previous buffer A, instead of being updated to B.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -1,4 +1,4 @@
-" Copyright (c) 2015 Junegunn Choi
+" Copyright (c) 2015-2019 Junegunn Choi
 "
 " MIT License
 "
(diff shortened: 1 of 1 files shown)


junegunn:
Works for me. Have you tested it with the minimal configuration?

https://raw.githubusercontent.com/junegunn/fzf.vim/master/.github/ISSUE_TEMPLATE.md

noboruma:
Outch.. it is working fine with the minimal configuration.
Sorry for all the fuss.

My config had the following:
`au BufWinEnter * if &fdm == 'expr' | setlocal foldmethod=manual | normal zM | endif`

Not entirely sure why this was responsible for the issue.
I am closing the PR anyway, thanks for your time!

PR #804, opened 2019-04-24
Add support for ctrl-a/ctrl-d in grep command

froody:
ctrl-a/ctrl-d are easier to use in a default tmux environment.


junegunn:
I need the original CTRL-A and I don't want to change the default people have been using for years out of sudden. You can override the bindings by passing `--bind` option to `fzf#vim#grep`.

```vim
call fzf#vim#grep('ag 1', 0, {'options': '--bind ctrl-a:select-all'})
```

PR #810, opened 2019-05-12
Workaround for last-position-jump in :Tags

liskin:
If there's an autocmd for BufReadPost that jumps to the last position,
searching for the tag location sometimes results in vim printing the
wrapscan warning and then requires hit-enter. Jump to the beginning
before searching for the tag to ensure consistent (no hit-enter)
behaviour.


junegunn:
How about `silent execute excmd`?

liskin:
@junegunn Oh, that's probably much better. :-)

junegunn:
Okay, please update the PR.

liskin:
Yep, I wanted to test it a bit more, it works very well, PR is updated. Thanks for the suggestion, this is indeed cleaner. :-) :+1:

junegunn:
Merged, thanks!

Issue #453, opened 2017-09-26
Is it possible to ignore some files


entropitor:
E.g. to ignore build files in Files, Ag, ...

junegunn:
Yes, https://github.com/junegunn/fzf#respecting-gitignore-hgignore-and-svnignore

strikemike2k:
> Yes, https://github.com/junegunn/fzf#respecting-gitignore-hgignore-and-svnignore

Updated link: https://github.com/junegunn/fzf#respecting-gitignore

mrded:
I suppose the question was not how to respect `.gitignore` but how to ignore some files in Files? 

Can we please reopen this?

bendoh:
@mrded please try this change https://github.com/junegunn/fzf.vim/pull/819 which will allow you to set `g:fzf_files_options.source` to influence the fzf source command (i.e., `rg --files ...`)

bendoh:
The problem with setting `FZF_DEFAULT_COMMAND` is that it requires an environment variable to be set outside of the vim process and it affects _all_ commands run through fzf. #819 will allow a more exact options specification

bendoh:
Guess that wasn't the best way, https://github.com/junegunn/fzf.vim/pull/819#issuecomment-498476100

tomskopek:
The docs are really good and helpful but maybe a bit intimidating to some - at least they were to me, before I played with it a bit more.

 I set FZF to use fd as per the example in the readme. You will need to `brew install fd` if you don't have it already

In my `~/.zshrc`:
```export FZF_DEFAULT_COMMAND='fd --type f --hidden --follow --exclude .git'```

This makes fzf use the output of `fd --type f` (same as doing `fd --type f | fzf` in the terminal - try it out!)

Now to exclude folders/files there are actually several ways to do it. One can ignore certain folders with fd by setting up a .fdignore file (see: https://github.com/sharkdp/fd#excluding-specific-files-or-directories)

Then in my repo's root directory, I created a .fdignore file with 'build' in it, and other folders I wanted to ignore.

```$cat .fdignore
mobile/ios
mobile/android
build```

Now if I use FZF inside vim those folders will be ignored.

Hope this helps someone one day!

Thanks junegunn for the excellent work and documentation.

**Comment by fkotsian** [comment]
For ag try (will search individual occurrences, for files I suggest the wonderful `fd` implementation above!):

```
# in ~/.fzf.{bash,zsh}
[...]

PR #819, opened 2019-06-03
Update how options are extended

bendoh:
Allow dictionaries to be used to extend the options provided to the base FZF plugin.

This will allow us to set `g:fzf_files_options` and set a different source so that we can influence the input to FZF's files command, e.g.,

```
let g:fzf_files_options = {'source': 'rg --files --no-ignore-vcs'} 
```

Will use ripgrep which will use `.ignore` but not `.gitignore`

-> Fold together option dictionaries verbatim

-> Extend `.options` with a list

-> Append / prepend .options string when not a dictionary

This will allow us to specify options for just the `files` command, addressing issue #453 without having to tweak an environment variable, and it allows for more flexibility in other commands if desired.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -303,7 +303,7 @@ function! fzf#vim#files(dir, ...)
   endif
 
   let args.options = ['-m', '--prompt', strwidth(dir) < &columns / 2 - 20 ? dir : '> ']
-  let args = s:merge_opts(args, get(g:, 'fzf_files_options', []))
+  let args = s:merge_opts(args, copy(get(g:, 'fzf_files_options', [])))
 
   return s:fzf('files', args, a:000)
 endfunction
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -46,22 +46,27 @@ endif
 
 function! s:extend_opts(dict, eopts, prepend)
   if empty(a:eopts)
-    return
+    return a:dict
   endif
-  if has_key(a:dict, 'options')
-    if type(a:dict.options) == s:TYPE.list && type(a:eopts) == s:TYPE.list
+
+  let extended = type(a:eopts) == s:TYPE.dict ? extend(a:eopts, a:dict) : a:dict
+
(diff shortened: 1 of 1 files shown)


junegunn:
`g:fzf_files_options` is deprecated and was [removed from the documentation two years ago](https://github.com/junegunn/fzf.vim/commit/2eaff049464e7b8304401dd4d79c86a4b0c4ed6c). The code is still there only for backward compatibility and it's not a recommended way to customize the behavior of the command. And you know, none of the other commands in fzf.vim understands `g:fzf_XXX_options`, so it was an exception.

The recommended way is to redefine each command using the autoload function behind it. README page shows many examples: https://github.com/junegunn/fzf.vim#advanced-customization.

```vim
command! -bang -nargs=? -complete=dir 
  \ Files call fzf#vim#files(<q-args>, {'source': 'rg --files --no-ignore-vcs'}, <bang>0)
```

This way, we don't have to maintain two dozens of extra configuration variables, and it's far more flexible (e.g. you can have multiple variations of the command at the same time with different names), though it can be a little daunting to the users who don't have any experience with VimScript.

```vim
command! ProjectFiles call fzf#vim#files('~/project')
command! TmpFiles call fzf#vim#files('/tmp', {'left': '30%', 'options': '--inline-info'})
```

Having said that, I don't see why you do not want to set `FZF_DEFAULT_COMMAND`. It's only used when fzf is started without standard input source.

```sh
export FZF_DEFAULT_COMMAND='echo hello'

# The variable doesn't really matter in most cases
ls | fzf
ps -ef | fzf
seq 1000 | fzf
```

And if you don't want it outside Vim for some reason, you can even define it in your Vim configuration file like so:

```vim
let $FZF_DEFAULT_COMMAND = '...'
```

dtantsur:
> And if you don't want it outside Vim for some reason, you can even define it in your Vim configuration file like so:

Does not work unfortunately... I wish something so important would be possible without modifying system-wide configuration :(

Actually, exporting this variable in the same shell you're running vim does not work either. Maybe it only works for Files, but not GFiles or Ag?

Issue #393, opened 2017-06-20
Ignore buffers in buffer list based on regex


ashiklom:
- Category
    - [ ] Question
    - [ ] Bug
    - [X] Suggestion
- OS
    - [ ] Linux
    - [ ] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [ ] Vim
    - [ ] Neovim

I would like to request the ability to ignore buffers based on a regular expression when using the `Buffer` command.

The implementation seems really straightforward -- just a two-line fix to the `s:buflisted()` function. For instance, to ignore terminal buffers:

```vim
function! s:buflisted()
  return filter(range(1, bufnr('$')), 'buflisted(v:val) && getbufvar(v:val, "&filetype") != "qf" && bufname(v:val) !~ g:fzf_ignore_string')
endfunction
```

...if `g:fzf_ignore_string is set.

JoshMcguigan:
To make it easier for others who might search for this, I was hoping to use this functionality to hide the `!/bin/bash` entry in the buffer list when using the `:terminal`.

bit101:
Very nice. I just hard coded it with `^term:.*` so it will stop showing my terminal buffer in the buffer list. Would be nice to have a way to do this. I will follow the pull request mentioned above. But in the meantime, this solves my current problem.

activaigor:
looks like a good solution for a problem, i wonder why it's not merged yet

PR #831, opened 2019-06-21
Enable configuration of the function used to retrieve the buffer list

s1341:
It's useful to enable configurability of the function used to retrieve the buffer list, so that certain buffers can be filtered, for example.

This pull request adds a new global variable which may contain the name of the function which provides the buffer list, allowing for extensibility of fzf's buffer related commands.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -193,8 +193,14 @@ for s:color_name in keys(s:ansi)
         \ "endfunction"
 endfor
 
+function! s:default_buffer_function()
+  return range(1, bufnr('$'))
+endfunction
+
 function! s:buflisted()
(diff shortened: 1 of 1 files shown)


vasilakisfil:
yes that's super helpful, any chance it gets accepted?

s1341:
Seeing as I have not received any communication from the maintainer, I think it unlikely. Still useful.

vasilakisfil:
This would also solve #393. @junegunn any thoughts ?

krkhan:
+1, I'm currently using this as a custom patch on top of `master`. @junegunn any concerns with merging this one?

s1341:
Any chance we could merge this?

bit101:
I've been patching my fzf install to add this functionality for over a year now. Would love to see this merged if at all possible.

PR #836, opened 2019-07-02
Bugfix: editing a command from history when colon is remapped

unclechu:
Added `'n'` (to not remap keys) argument to `feedkeys` function.

I personally have `:` key remapped and `ctrl-e` doesn't work correctly for `:History:`.


junegunn:
Thanks!

PR #891, opened 2019-10-31
Change &columns to winwidth(0)

monkoose:
It will support floating and side windows(if it lower then 120 characters). Because `&columns` return
screen size and not window size. And actually do nothing for such windows.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -306,7 +306,7 @@ function! fzf#vim#files(dir, ...)
     let dir = s:shortpath()
   endif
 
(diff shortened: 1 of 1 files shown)


junegunn:
The code runs before the window for fzf is open, so I don't think this solves the problem.

monkoose:
It works for me. For things where it used like FzfLines it doesn't show filenames only buffer number if window is < then s:wide that is 120 columns.

monkoose:
Yes. You are right.
But I'm 100% sure i tested it with different windows and it worked. And now i testing and it doesn't. Seems like i tested it wrong previously.

PR #892, opened 2019-10-31
Fix preview of paths starting with ~/

ptzz:
Fixes outstanding comments in #774 and #890.

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -17,7 +17,7 @@ if [[ $1 =~ ^[A-Z]:\\ ]]; then
   CENTER=${INPUT[2]}
 fi
 
-FILE="${FILE/#\~/$HOME}"
+FILE="${FILE/#\~\//$HOME\/}"
 if [ ! -r "$FILE" ]; then
   echo "File not found ${FILE}"
   exit 1
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -17,6 +17,7 @@ if [[ $1 =~ ^[A-Z]:\\ ]]; then
   CENTER=${INPUT[2]}
 fi
 
+FILE="${FILE/#\~/$HOME}"
 if [ ! -r "$FILE" ]; then
   echo "File not found ${FILE}"
   exit 1


junegunn:
Thanks.

tzhouhc:
On zsh when converting FILE this operation adds an extra backslash after the $HOME:

```
minecraft@nookie:~$ FILE='~/.dotfiles'
minecraft@nookie:~$ FILE="${FILE/#\~\//$HOME\/}"
minecraft@nookie:~$ echo $FILE
/home/minecraft/.dotfiles
minecraft@nookie:~$ zsh
nookie% FILE='~/.dotfiles'
nookie% FILE="${FILE/#\~\//$HOME\/}"
nookie% echo $FILE
/home/minecraft\/.dotfiles
```

Causing history preview to, instead of failing because it doesn't know about ~ in single quotes, fail because it doesn't know why there's a backtick in the middle:

```
File not found /Users/tingzhou\/.dotfiles/vim/testrc.vim
```

Setting shell in vim doesn't seem to help.

attached, minimal testing vimrc:

```vimL
let vimDir = '$HOME/.vim'
let &runtimepath.=','.vimDir

set nocompatible
call plug#begin('~/.vim/bundle')

" fzf -- quick jump to file, tag and such
Plug 'junegunn/fzf', { 'dir': '~/.fzf' }
Plug 'junegunn/fzf.vim'

call plug#end()"}}}

" History preview with colorings and such
command! -bang -nargs=* History
  \ call fzf#vim#history(fzf#vim#with_preview(), <bang>0
  \ )

nnoremap <c-p> :History<cr>
```

Updating the line to

```sh
FILE="${FILE/#\~/$HOME/}"
```

seems to fix, and testing manually on the shell doesn't *seem* to break it either, but I do not know for sure.

tzhouhc:
Seeing the history, I think if ensuring we only substitute "~/" instead of "~", the solution would be to use

```bash
FILE="${FILE/#\~\//$HOME/}"
```

junegunn:
> On zsh

It's [a bash script](https://github.com/junegunn/fzf.vim/blob/master/bin/preview.sh#L1).

tzhouhc:
> > On zsh
> 
> It's [a bash script](https://github.com/junegunn/fzf.vim/blob/master/bin/preview.sh#L1).

I feel stupid.

But now I am out of ideas about why it behaves this way. I wonder if there's some obscure env var that changes the substitution rules, since while this (now) is obviously not a bash/zsh thing, the issue remains that I am getting an extra `\` :(

tzhouhc:
> > On zsh
> 
> It's [a bash script](https://github.com/junegunn/fzf.vim/blob/master/bin/preview.sh#L1).

> > On zsh
> 
> It's [a bash script](https://github.com/junegunn/fzf.vim/blob/master/bin/preview.sh#L1).

I feel stupid.

But now I am out of ideas about why it behaves this way. I wonder if there's some obscure env var that changes the substitution rules, since while this (now) is obviously not a bash/zsh thing, the issue remains that I am getting an extra `\` :(

junegunn:
Does this correctly print bash version? `/usr/bin/env bash --version`

tzhouhc:
Anyway, solved by installing bash 5 and having it on my path before the mac included one. If anyone somehow gets this same issue, `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin19)` (the one on Catalina) is the culprit. This behavior is not present on darwin18.

Thanks to @junegunn for helping out!

PR #911, opened 2019-11-27
Add searches to vim search history, even without Ctrl-E.

infokiller:


infokiller:
Friendly ping, can this PR be merged please? Thanks!

junegunn:
I don't understand what this is about. The selected item is automatically added to history. Related: https://github.com/junegunn/fzf.vim/issues/661

infokiller:
Hmm, looking back I'm also not sure what that tried to fix... I suspect what happened is that I used to have a mapping for `/`, and because `g:__fzf_command` used `normal` and not `normal!`, it didn't add it to the history.
I'll send a different PR to change `normal` to `normal!`.

Issue #918, opened 2019-12-17
BCommit doesn't show history of file in submodule


matawed:
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

If you do `:BCommit` on a buffer which is located in a git submodule, but the current working directory is in main git clone, then the history of the main git repository is shown instead of the submodule. 

e.g.
 
```
git clone <repo>
cd <repo>
git submodule update --init --recursive
# file test.cpp is in submodule
vim submodule/test.cpp
``` 

if you enter `:BCommit` then you'll get the history of <repo> instead of submodule.

PR #919, opened 2019-12-17
Add support for files in submodules when using BCommits/Commits

matawed:
If file in buffer is located in submodule then bcommit doesn't show the history of this file.

fixes: https://github.com/junegunn/fzf.vim/issues/918

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1105,7 +1105,7 @@ function! s:commits(buffer_local, args)
     return s:warn('Not in git repository')
   endif
   let path = expand('%:p:h')
-  let source = 'git -C '.path.' log '.get(g:, 'fzf_commits_log_options', '--color=always '.fzf#shellescape('--format=%C(auto)%h%d %s %C(green)%cr'))
+  let source = 'git -C '.fzf#shellescape(path).' log '.get(g:, 'fzf_commits_log_options', '--color=always '.fzf#shellescape('--format=%C(auto)%h%d %s %C(green)%cr'))
   let current = expand('%')
   let managed = 0
   if !empty(current)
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -1104,20 +1104,19 @@ function! s:commits(buffer_local, args)
   if empty(s:git_root)
     return s:warn('Not in git repository')
   endif
-
-  let source = 'git log '.get(g:, 'fzf_commits_log_options', '--color=always '.fzf#shellescape('--format=%C(auto)%h%d %s %C(green)%cr'))
+  let path = expand('%:p:h')
```
junegunn: We should wrap it with `fzf#shellescape` for it to work on different platforms.

autoload/fzf/vim.vim
```diff
@@ -1104,20 +1104,19 @@ function! s:commits(buffer_local, args)
...
     call system('git show '.fzf#shellescape(current).' 2> '.(s:is_win ? 'nul' : '/dev/null'))
     let managed = !v:shell_error
   endif
-
   if a:buffer_local
     if !managed
       return s:warn('The current buffer is not in the working tree')
     endif
-    let source .= ' --follow '.fzf#shellescape(current)
+    let source .= ' --follow '.expand('%:t')
```
junegunn: `fzf#shellescape` is still needed here too.

matawed: Reverted change as it was not needed.

autoload/fzf/vim.vim
matawed: done.

junegunn:
Thanks, but I had to revert the change as `BCommits` stopped working for me. Have you tested it with a file on a repository without submodules?

matawed:
It seems my tests weren't deep enough. For files with some (one?) change(s) it works. See e.g. 

image

vs

image

Will check.

PR #928, opened 2020-01-09
(bugfix) filenames with double-colons will lead to erroneous behaviour

atarax:
Opening files with a double-colon results in wrong filename getting used for opening the file.
The fix is pretty naive and static, it replaces all double-colon with another (less-used) character and so far works only for exactly two repeating colons (i tried to make it a regex but it didn't work, i'm a vimscript rookie)

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -641,7 +641,7 @@ endfunction
 " Ag / Rg
 " ------------------------------------------------------------------
 function! s:ag_to_qf(line, has_column)
-  let parts = split(a:line, ':')
+  let parts = split(a:line, '[^:]\zs:\ze[^:]')
   let text = join(parts[(a:has_column ? 3 : 2):], ':')
   let dict = {'filename': &acd ? fnamemodify(parts[0], ':p') : parts[0], 'lnum': parts[1], 'text': text}
   if a:has_column


junegunn:
How about `split(a:line, '[^:]\zs:\ze[^:]')`?

atarax:
@junegunn :genius! what kind of witchcraft if this :D i assume there is special vim-related regex-logic? would be nice if you can help me understand and thanks a lot!

junegunn:
Yeah, it's a non-standard vim-specific regular expression (see `:help \zs`). It splits the line with a colon preceded and followed by any character that is not a colon, and those surrounding characters are not included in the matching string.

junegunn:
Merged, thanks!

atarax:
Thanks a lot :)

wis:
@junegunn you should start using ripgrep's --null option, it would propably save you the headache of these types of bugs that I found many of browsing the issues.
```
       -0, --null
           Whenever a file path is printed, follow it with a NUL byte. This includes printing
           file paths before matches, and when printing a list of matching files such as with
           --count, --files-with-matches and --files. This option is useful for use with
           xargs.
```
demo:
testripgrepnul.py
```.py
#!/usr/bin/python3
import sys
for line in sys.stdin.readlines():
    print(line.split("\0"))
```
$ ` rg --vimgrep --null --no-messages --max-depth 1 --hidden face | python testripgrepnul.py`
```
['.emoji', '1:15:😀 grinning face; U+1F600\n']
['.emoji', '2:15:😃 grinning face with big eyes; U+1F603\n']
['.emoji', '3:15:😄 grinning face with smiling eyes; U+1F604\n']
['.emoji', '4:14:😁 beaming face with smiling eyes; U+1F601\n']
['.emoji', '5:25:😆 grinning squinting face; U+1F606\n']
['.emoji', '6:15:😅 grinning face with sweat; U+1F605\n']
['.emoji', '8:6:😂 face with tears of joy; U+1F602\n']
['.emoji', '9:23:🙂 slightly smiling face; U+1F642\n']
['.emoji', '10:18:🙃 upside-down face; U+1F643\n']
['.emoji', '11:14:😉 winking face; U+1F609\n']
['.emoji', '12:14:😊 smiling face with smiling eyes; U+1F60A\n']
['.emoji', '13:14:😇 smiling face with halo; U+1F607\n']
['.emoji', '14:14:🥰 smiling face with 3 hearts; U+1F970\n']
```
it splits off the path (which can have any character but the null character on linux) from the position+the line

PR #933, opened 2020-01-14
[[B]Commmits] Add support for files in submodules

matawed:


matawed:
follow-up of https://github.com/junegunn/fzf.vim/pull/919

matawed:
ok, one problem for me was that if you open vim like `vim path/to/file` and then do `:BCommits` there, then you got an empty BCommits pane, this is fixed now. Did that fail for you or what did you test?

matawed:
For one thing I need your help, though. I you open a file in vim with 
`vim /path/to/file` and do `;BCommits` in there. The BCommits pane with the commit diff shows `unknown revision or path not in working tree`. History pane looks fine though. E.g.

image

If you navigate to that directory in advance and only do `vim file` followed by `:BCommits` in there it works fine. So the problem is that we have to do _fzf_ in that directory rather than in current working dir. Is there an option for that you do we have to change cwd (and go back afterwards again)?

Issue #182, opened 2016-08-01
Reuse previous search/query


kevintraver:
- Category
    - [x] Question
    - [ ] Bug
    - [ ] Suggestion
- OS
    - [x] Linux
    - [x] Mac OS X
    - [x] Windows
    - [x] Etc.
- Vim
    - [x] Vim
    - [x] Neovim

Unite.vim had functionality to bring up the previous used search/query.

https://github.com/Shougo/unite.vim/blob/06b927532e1a92a948a011a07f4eb799bc8710ba/doc/unite.txt#L352-L359

I was wondering if this or something similar is possible in fzf.vim?

junegunn:
I'm not completely sure what the command of Unite does, but you can enable search history by setting `g:fzf_history_dir`. See the description in https://github.com/junegunn/fzf.vim#global-options

kevintraver:
The command UniteResume would bring up your previous unite command, with the previous search already populated. 

Example:
- `:Ag`
- type in _fasdfasdf_
- hit **enter**

I was looking if there was a way to then bring up `:Ag` with the most recent search (in this case _fasdfasdf_)

NovaDev94:
@kevintraver With `g:fzf_history_dir` you can

* `:Ag`
* type in `something`
* hit *enter*
* `:Ag` again
* `Ctrl+P/N`
* Your most recent search is there

NovaDev94:
@junegunn One small problem (not sure if it's actually a problem) with fzf.vim is I can't use the search history when I'm using some commands like `:Ag my_word_here`. Seem like fzf only stores the query string input in the fzf window, not the one sent as argument to the command.  

Is there anyway to overcome this? I'm having this binding

```vim
nnoremap K :Ag <C-r><C-w><cr>
```

junegunn:
@NovaDev94 

> Seem like fzf only stores the query string input in the fzf window, not the one sent as argument to the command.


Yeah, fzf is a filter and does not know about the command used to populate the list. One workaround is to pass the query string to fzf using `-q` option, so that the string is stored in the history directory.

```vim
command! -nargs=* AgQ call fzf#vim#ag(<q-args>, {'down': '40%', 'options': '-q '.shellescape(<q-args>.' ')})
```

NovaDev94:
Thanks alot @junegunn, you made such a great tool!

edi9999:
This can be closed

rohanghige:
[...]

PR #941, opened 2020-02-02
Add `Resume` command

artemave:
### What

A command that opens last used fzf command buffer (with the query).

Example mapping:

```vimscript
nnoremap <leader>u :Resume<cr>
```

### Why

Because I often run the same query. And I am missing `UniteResume`

cc #182

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -200,7 +200,13 @@ endfunction
 let s:last_command = {}
 
 function! s:fzf(name, opts, extra)
-  let s:last_command = {'name': a:name, 'opts': a:opts, 'extra': a:extra, 'grep_command': $FZF_DEFAULT_COMMAND}
+  let s:last_command = {
+        \ 'name': a:name,
+        \ 'opts': a:opts,
+        \ 'extra': a:extra,
+        \ 'time': localtime(),
+        \ 'grep_command': $FZF_DEFAULT_COMMAND
+        \}
 
   let [extra, bang] = [{}, 0]
   if len(a:extra) <= 1
@@ -219,7 +225,6 @@ function! s:fzf(name, opts, extra)
   let eopts  = has_key(extra, 'options') ? remove(extra, 'options') : ''
   let merged = extend(copy(a:opts), extra)
   call s:merge_opts(merged, eopts)
-
   return fzf#run(s:wrap(a:name, merged, bang))
 endfunction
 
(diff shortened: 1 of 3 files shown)


junegunn:
Thanks, but I don't want to merge this. You could just write a general-purpose command wrapper and use it.

```vim
function! s:remember(command)
  execute a:command
  " Or save it to a file for persistence
  let g:prev_command = a:command
endfunction

command! -nargs=* Remember call s:remember(<q-args>)
command! Redo execute g:prev_command
```
Then you can `:Remember Ag foo` and `:Redo`. You can use it with any commands with any arguments.

artemave:
This is interesting - it didn't occur to me that it could be a command wrapper. One thing is not clear though: how would I go about restoring fzf `--query` in this case?

artemave:
@junegunn how would I go about restoring fzf query in case of a wrapper?

Using `<c-p>` does not work because it picks up a query from the last command _with a query_, and that may not have necessarily been the last command.

junegunn:
So if I understood correctly, you want to maintain separate query history for each command with different arguments. i.e. `Ag foo`, `Ag bar`, ...

Sorry but it's not something I want to add to this project. I personally never needed such a feature (actually I don't even use `g:fzf_history_dir`), and I don't like that it'll create a bunch of history files.

Note that you can manually override `--history` option:

```vim
let g:fzf_history_dir = '/tmp/fzf-history'

echo fzf#wrap('type1').options
echo fzf#wrap('type1', { 'options': ['--history', '/tmp/fzf-history/type2'] }).options
```

So you can write your own version of each command,

```vim
command! -bang -nargs=* Rg call fzf#vim#grep("rg --column --line-number --no-heading --color=always --smart-case ".shellescape(<q-args>), 1, {'options': ['--history', g:fzf_history_dir.'/rg-'.<q-args>]}, <bang>0)
```

and combine with `Remember` command shown above.

artemave:
> So if I understood correctly, you want to maintain separate query history for each command with different arguments.

Well, fzf history (or perhaps fzf.vim?) creates a different query history file for each command (including custom commands) already:

```
% ls ~/.fzf-history/
buffers  files  gfiles-diff  history-files  lines  marks  rg  tags
```

I simply take advantage of that feature to look up the last query for a command (I do keep track of a last command, but there is no way to know the actual query since it happens inside fzf itself).

artemave:
Update: I am using `Resume` _all_ the time. I also keep my fork up-to-date. In case you ever change your mind!

artemave:
Update: improved the implementation such that it now remembers the last query even if the user didn't press enter. That's useful when browsing through files with preview without actually opening any.

PR #962, opened 2020-03-11
Fix(preview.sh): Fix Regex that replace tilde in wrong way

ryuta69:
The regex is supposed to replace tilde to $HOME.

However the regex used unnecessary escaping, so the replacing was
mistaken.

The path always showed "File not found" in preview window.

```bash
FILE='~/work/test.sh'

FILE_before_fix="${FILE/#\~\//$HOME\/}"
echo $FILE_before_fix
# FILE_before_fix = /Users/user\/work/test.sh
# unnecessary escaping has shown up.

FILE_after_fix="${FILE/#\~\//$HOME/}"
echo $FILE_after_fix
# FILE_after_fix = /Users/user/work/test.sh
# This works right.
```

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -17,7 +17,7 @@ if [[ $1 =~ ^[A-Z]:\\ ]]; then
   CENTER=${INPUT[2]}
 fi
 
-FILE="${FILE/#\~\//$HOME\/}"
+FILE="${FILE/#\~\//$HOME/}"
 if [ ! -r "$FILE" ]; then
   echo "File not found ${FILE}"
   exit 1


junegunn:
I can't reproduce the problem. And since this is a bash script, make sure to test it on bash.

https://github.com/junegunn/fzf.vim/blob/6f1599cc857fa7e3f7fc0bfe927dd968d5c1f553/bin/preview.sh#L1

junegunn:
```sh
> echo $BASH_VERSION
5.0.11(1)-release

> FILE='~/work/test.sh'

> echo "${FILE/#\~\//$HOME\/}"
/Users/jg/work/test.sh

> echo "${FILE/#\~\//$HOME\/\/}"
/Users/jg//work/test.sh

> echo "${FILE/#\~\//$HOME/}"
/Users/jg/work/test.sh

> echo "${FILE/#\~\//$HOME//}"
/Users/jg//work/test.sh
```

But you're right that the escaping is unnecessary here, I'll merge this. Thanks.

Issue #997, opened 2020-04-26
Show filename in Maps comment as well as line number


ianhomer:
Currently `:Maps` shows a number at the end the line which is the line in the file that the map was defined.

![Screenshot 2020-04-26 at 17 13 00](https://user-images.githubusercontent.com/190819/80313229-7c3d3980-87e1-11ea-9409-609f5c3a49b7.png)

It'd be useful if the file name could be shown as well to give more context on where it was configured.

(I'll post PR shortly for this change)

ianhomer:
After apply fix in PR results look like

![Screenshot 2020-04-26 at 17 19 48](https://user-images.githubusercontent.com/190819/80313369-785de700-87e2-11ea-83ce-899483bef6f6.png)

Issue #324, opened 2017-02-26
cannot check the mappings


imsuwj:
- Category
    - [ ] Question
    - [x] Bug
    - [ ] Suggestion
- OS
    - [x] Linux
    - [ ] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [x] Vim
    - [ ] Neovim

VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Feb 14 2017 10:57:50)

![screenshot_20170226_133934](https://cloud.githubusercontent.com/assets/2463831/23337450/17e71aba-fc29-11e6-9f54-7c2f17b25e41.png)

When i using command `:Maps`,it shows me `recently modified /usr/share/vim/vim80/plugin/netrwPlugin.vim`

imsuwj:
And i found it is related to the locale

this is the locale in KDE,Chinese interface
```
LANG=zh_CN.UTF-8
LC_CTYPE=en_US.UTF-8
LC_NUMERIC="zh_CN.UTF-8"
LC_TIME="zh_CN.UTF-8"
LC_COLLATE="zh_CN.UTF-8"
LC_MONETARY="zh_CN.UTF-8"
LC_MESSAGES="zh_CN.UTF-8"
LC_PAPER="zh_CN.UTF-8"
LC_NAME="zh_CN.UTF-8"
LC_ADDRESS="zh_CN.UTF-8"
LC_TELEPHONE="zh_CN.UTF-8"
LC_MEASUREMENT="zh_CN.UTF-8"
LC_IDENTIFICATION="zh_CN.UTF-8"
LC_ALL=
```

Then if set them to `en_US.UTF-8`, the mappings seems normal
```
LANG=en_US.UTF-8`
LC_CTYPE=en_US.UTF-8
LC_NUMERIC="en_US.UTF-8`"
...
```

imsuwj:
But i run `echo $LANG` in vim : `en_US`

junegunn:
Hmm, maybe it's because Maps command parses the output of `verbose nmap`. It looks for `Last set from ...`.

```
n  <Space>ag   * :Ag <C-R><C-W><CR>
        Last set from ~/.vimrc
n  <Space><CR> * :Buffers<CR>
        Last set from ~/.vimrc
```

I guess `Last set from` is printed in Chinese in your case?

imsuwj:
@junegunn yes, then how to solve it

junegunn:
[...]

PR #998, opened 2020-04-26
#997 Extract filename and line number for maps comment

ianhomer:
Change to address https://github.com/junegunn/fzf.vim/issues/997

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1214,7 +1214,7 @@ function! fzf#vim#maps(mode, ...)
   let curr = ''
   for line in split(cout, "\n")
     if line =~ "^\t"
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, here's what I'm suggesting in addition to your patch:

```vim
let src = "\t".substitute(matchstr(line, '/\zs[^/\\]*\ze$'), ' [^ ]* ', ':', '')
```

- Backslash for Windows
- Tab instead of two spaces for slightly better alignment
- Replace `foo.vim line 100` to `foo.vim:100`
    - The actual string `line may differ depending on the locale setting (see #324)

ianhomer:
PR updated as suggested, looks like the following now on my Mac

Screenshot 2020-04-30 at 20 32 07

junegunn:
Merged, thanks.

Issue #657, opened 2018-06-19
Ripgrep support


ksantr:
- Category
    - [ ] Question
    - [ ] Bug
    - [x] Suggestion
- OS
    - [x] Linux
    - [x] macOS
    - [x] Windows
    - [x] Etc.
- Vim
    - [x] Vim
    - [ ] Neovim

For me ag is working very strangely, for really big project it's very slow. I prefer to use ctags, but I'm not sure that ctags in fzf regenerates new tags after updates. What about ripgrep? It's incredibly fast.

junegunn:
See `:Rg` example in https://github.com/junegunn/fzf.vim#advanced-customization

devth:
How do you use `rg` for `:Files` though? I can't seem to find the right `rg` options to make it work.

Tried:

```
export FZF_DEFAULT_COMMAND='rg --line-number --color never --no-heading ""'
```

thepenguinthatwants:
let $FZF_DEFAULT_COMMAND='rg --files'
Not sure if it really uses rg but I am using this.

thepenguinthatwants:
> How do you use `rg` for `:Files` though? I can't seem to find the right `rg` options to make it work.
> 
> Tried:
> 
> ```
> export FZF_DEFAULT_COMMAND='rg --line-number --color never --no-heading ""'
> ```

let $FZF_DEFAULT_COMMAND='rg --files'
Not sure if it really uses rg but I am using this.

PR #1001, opened 2020-04-28
Fix(vim.vim): Error calling Files when ripgrep is FZF_DEFAULT_COMMAND

ryuta69:
related to https://github.com/junegunn/fzf.vim/issues/657 .

When users set FZF_DEFAULT_COMMAND as `rg` or `rg ~~~`, `Files` command fails.  It is because it calls raw `fzf` which is same as FZF_DEFAULT_COMMAND, but command is supposed to be 'rg --files' to look up files under current directory.

I came up with thinking that I need to fix fzf side, but [`fzf#vim#grep` has same approach](https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L737-L743) replacing FZF_DEFAULT_COMMAND, so I make PR in fzf.vim side.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -323,7 +323,16 @@ function! fzf#vim#files(dir, ...)
 
   let args.options = ['-m', '--prompt', strwidth(dir) < &columns / 2 - 20 ? dir : '> ']
   call s:merge_opts(args, get(g:, 'fzf_files_options', []))
-  return s:fzf('files', args, a:000)
+
+  try
+    let prev_default_command = $FZF_DEFAULT_COMMAND
+    if !empty(matchstr(prev_default_command, '^rg'))
+      let $FZF_DEFAULT_COMMAND = 'rg --files'
+    endif
+    return s:fzf('files', args, a:000)
+  finally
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but I don't quite see the point of the pull request. It's up to users to use a correct FZF_DEFAULT_COMMAND. We can't fix it for them as they can use different variations of `rg` command. e.g. `rg --files --hidden`, `rg --files --hidden --follow`, etc.

The reason `fzf#vim#grep` sets FZF_DEFAULT_COMMAND is very different. It is for `reload` action. See https://github.com/junegunn/fzf/issues/1750 for details.

ryuta69:
@junegunn 

Thank you for your reply, but this happens even when user sets `FZF_DEFAULT_COMMAND = 'rg'` without any options.

It's because ripgrep needs to call 'rg --files' to search files under current directory, however, fzf calls only 'rg' which is just FZF_DEFAULT_COMMAND right now. Therefore I thought it would be useful to call 'rg --files' only when user uses :File.

junegunn:
> when user sets FZF_DEFAULT_COMMAND = 'rg' without any options.

That's clearly a misconfiguration. The user, not we, should fix it.

ryuta69:
okey, thank you.

PR #1007, opened 2020-05-03
add fzf-complete-file-rg

cassiobotaro:
As we have (fzf-complete-file-ag), I add (fzf-complete-file-rg) to complete files using rg instead of ag.


junegunn:
Thanks, but I think I'm going to deprecate `fzf-complete-file-ag` and remove it from the documentation, as we have a more flexible way to achieve the same.

junegunn:
See https://github.com/junegunn/fzf.vim/commit/5b6f282d97e7ce732d39baff2d58d89629b09385

Issue #988, opened 2020-04-17
Preview doesn't work on Windows


habamax:
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

Instead of a file preview I get `/bin/bash: .... no such

![image](https://user-images.githubusercontent.com/234774/79543533-4dfb8380-8096-11ea-9d6c-cee8890ee5ab.png)

habamax:
Looks like bash on windows can't use absolute paths?

![image](https://user-images.githubusercontent.com/234774/79544325-851e6480-8097-11ea-9b0b-d8f107059ed8.png)

habamax:
BTW, I saw a closed issue which suggests that preview would use ruby script if ruby is available.

I have it available and still fzf tries to use bash

michael-angelozzi:
I have the same issue, except for me the preview always says:
```
Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore 
```

janlazo:
@michael-angelozzi That is WSL bash, not git bash.

michael-angelozzi:
@janlazo  On my machine:
```
Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
Press any key to continue...
```
The topic "Preview doesn't work on Windows" is the same issue for me, that's why I posted in this thread instead of starting a new.

janlazo:
It finds WSL bash first so adjust your `PATH` environment variable so that cmd.exe or vim/nvim finds git bash or any working bash executable (ie. busybox) first in PATH.

I suppose fzf and fzf.vim can fail early if this happens but that won't close the issue.

GasparVardanyan:
Nothing helped me so I've launched cmd as administrator and deleted C:\System32\bash.exe :d
Now vim-fzf works in my nvim !!

PR #1009, opened 2020-05-03
Fix preview bin directory for git bash

ladevieq:
Hey thanks for the great work on fzf.

This PR is supposed to fix https://github.com/junegunn/fzf.vim/issues/988.
If you have suggestions, i can edit the code and test on my Windows computer.

The preview bin is not found on window using git for windows.
The issue is that git bash is using a unix like path architecture
(disk mounted in /mnt dir).

The fix replace, path disk name with unix path to disk style, and `\`
with `/`.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -28,8 +28,6 @@ set cpo&vim
 " Common
 " ------------------------------------------------------------------
 
-" 'preview': '/mnt/c/Users/ladev/vimfiles/pack/default/start/fzf.vim/bin/preview.sh',
-
 let s:is_win = has('win32') || has('win64')
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:bin_dir = expand('<sfile>:h:h:h').'/bin/'
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -28,6 +28,8 @@ set cpo&vim
 " Common
 " ------------------------------------------------------------------
 
+" 'preview': '/mnt/c/Users/ladev/vimfiles/pack/default/start/fzf.vim/bin/preview.sh',
+
 let s:is_win = has('win32') || has('win64')
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:bin_dir = expand('<sfile>:h:h:h').'/bin/'
@@ -41,7 +43,8 @@ if s:is_win
   else
     let s:bin.preview = fnamemodify(s:bin.preview, ':8')
   endif
-  let s:bin.preview = 'bash '.escape(s:bin.preview, '\')
+    let git_bash_bin = 'bash /mnt/'.substitute(s:bin.preview, '\([A-Z]\):', '\L\1\e', '')
+    let s:bin.preview = substitute(git_bash_bin, '\', '/', 'g')
 endif
 
 let s:wide = 120


habamax:
I was skeptical at first... but it works for me. Thx!

image

junegunn:
@janlazo Hi, sorry to bother you but do you think this is a proper fix?

autoload/fzf/vim.vim
```diff
@@ -41,7 +41,8 @@ if s:is_win
   else
     let s:bin.preview = fnamemodify(s:bin.preview, ':8')
   endif
-  let s:bin.preview = 'bash '.escape(s:bin.preview, '\')
+    let git_bash_bin = 'bash /mnt/'.substitute(s:bin.preview, '\([A-Z]\):', '\L\1\e', '')
```
janlazo, line 44: Why is `/mnt/` here? What if git with another method? What if the bash executable is not git bash?

ladevieq: The best fix would probably be to write a Powershell script that would be use for windows.
This would require to bypass the execution policy, [see](https://riptutorial.com/powershell/example/20107/bypassing-execution-policy-for-a-single-script).

If you are ok with this, i can create another PR with a Powershell script.

habamax:
Something that doesn't require additional tools to be installed would be nice.
I wonder if it is possible to use plain bat/cmd files or built in js?

janlazo:
What is the shell set in the vimrc, shell options for that shell, and the shell (environment) used to run vim/nvim?

Windows allows forward slash for running binaries via executable path but bash and git need to be executed directly without the user's shell. Preview commands was coded to make it run a temporary batchfile which runs cmd.exe so changing to forward slash can only make things worse.

autoload/fzf/vim.vim
```diff
@@ -41,7 +41,8 @@ if s:is_win
...
+    let s:bin.preview = substitute(git_bash_bin, '\', '/', 'g')
```
janlazo, line 45: If you want to use mixed paths without using cygpath, ok. Problem is if it breaks the shortpath for Vim.

habamax:
> 
> 
> What is the shell set in the vimrc, shell options for that shell, and the shell (environment) used to run vim/nvim?

All are default in both vim and nvim, what exact option names you have in mind?

shell? shellquote? shellxquote? shellredir? 

> Windows allows forward slash for running binaries via executable path but bash and git need to be executed directly without the user's shell. Preview commands was coded to make it run a temporary batchfile which runs cmd.exe so changing to forward slash can only make things worse.

But... it directly runs bash with preview.sh as a parameter, did I miss anything?

autoload/fzf/vim.vim
janlazo: I'm ok with it but the powershell script should be cross-platform if possible so that it can be tested on Linux.

janlazo:
> I wonder if it is possible to use plain bat/cmd files or built in js?

Best to go with powershell if we're considering a different language that must work on Windows. batchfiles are hard to maintain and JS is not available by default.

janlazo:
Preview command should work with vim/nvim defaults especially if the shell options like shellcmdflag and shellslash are left at their default values. It's unusual for it to fail unless there is a regression.

IIRC, fzf-vim passes a preview command to fzf (binary). fzf parses the preview command for a placeholder but the parsing logic is for `sh` shells. Then, fzf passes the parsed command to the shell. It is always cmd.exe on Windows because `SHELL` environment variable is no-opt so backward slashes are used for filepaths and passing shortpaths to cmd.exe avoids escaping issues. Using mixed paths and maybe double quotes for shell escape could be sufficient if there is no parsing issue from fzf (binary).

`/mnt/` should not be required because git bash has access to the network drives. On my system, sh and bash do not have `/mnt/`.

habamax:
> Preview command should work with vim/nvim defaults especially if the shell options like shellcmdflag and shellslash are left at their default values. It's unusual for it to fail unless there is a regression.

Everything is default. And if you look into the issue I have posted results of plain bash calls with preview.sh as parameter. It just doesn't work with full path to preview.sh. https://github.com/junegunn/fzf.vim/issues/988#issuecomment-615094381

I suspect the issue is with WSL -- I have it installed and it (probably?) provides bash.exe in system32/ directory. And this bash.exe doesn't work with `C:\path\to\preview.sh`

habamax:
Indeed the issue is with WSL:
image

It does install it in system32 folder and bash from there is found before other bashes.

janlazo:
Is it enough to recommend the user to adjust their PATH so wsl bash is not found first?
fzf.vim can add a warning and exit early or adjust the preview command to use `/mnt/${network_drive}/` if wsl bash is found first.

Does WSL bash have a unique text in the output of `--version` or `--help` so fzf.vim can parse it?

habamax:
> Does WSL bash have a unique text in the output of `--version` or `--help` so fzf.vim can parse it?

Not sure
```
PS C:\Users\maksim.kim> bash --version
GNU bash, version 5.0.3(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

```

PS C:\Users\maksim.kim> bash --help
GNU bash, version 5.0.3(1)-release-(x86_64-pc-linux-gnu)
Usage:  /bin/bash [GNU long option] [option] ...
        /bin/bash [GNU long option] [option] script-file ...
GNU long options:
        --debug
        --debugger
        --dump-po-strings
        --dump-strings
        --help
        --init-file
        --login
        --noediting
        --noprofile
        --norc
        --posix
        --pretty-print
        --rcfile
        --restricted
        --verbose
        --version
Shell options:
        -ilrsD or -c command or -O shopt_option         (invocation only)
        -abefhkmnptuvxBCHP or -o option
Type `/bin/bash -c "help set"' for more information about shell options.
Type `/bin/bash -c help' for more information about shell builtin commands.
Use the `bashbug' command to report bugs.

bash home page: <http://www.gnu.org/software/bash>
General help using GNU software: <http://www.gnu.org/gethelp/>
```

habamax:
> Is it enough to recommend the user to adjust their PATH so wsl bash is not found first?

I don't think so. It would be quite cumbersome -- my git and other tools paths are managed by scoop for example. I am not sure if wsl would work if I I put wsl bash path after git bash.

habamax:
> Does WSL bash have a unique text in the output of `--version` or `--help` so fzf.vim can parse it?

What I know: bash in wsl has `$WSLENV` set


according to https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash you can check it in bash with:

    if [ -z ${WSLENV+x} ]; then echo "WSLENV is unset"; else echo "var is set to '$WSLENV'"; fi

I don't know how to invoke it from command line though.

PR #1012, opened 2020-05-05
use -- before the search pattern for Ag, Rg, etc.

kAworu:
Before this patch, a search pattern starting with a dash like
`:Rg -bang` would fail (or exhibit the wrong behaviour) because it would
be treated as a option. However, this case is very common when searching
for ->member in a project in C, C++, PHP, etc.


junegunn:
Merged, thanks!

Issue #1013, opened 2020-05-06
:CwdHistory implementation?


jesseleite:
Not sure if you'd consider implementing `:CwdHistory` natively in fzf.vim here?  I've implemented it locally ([see my dotfiles commit here](https://github.com/jesseleite/dotfiles/commit/9d4745479e69ed86327e5ed6449eb73c044dd8ee)) and it works well.  Basically, I just needed to filter by this custom `s:file_is_in_cwd()` function...

```vim
function! s:file_is_in_cwd(file)
  return filereadable(a:file) && match(a:file, getcwd() . '/') == 0
endfunction
```

The rest of the code is copy and paste, because so many of your functions are `s:` script local, so it's hard to extend locally without copying a whack of stuff.  Anyway, thoughts?

---

- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

jesseleite:
Or maybe even `:GHistory` (git files history), kind of like what `:GFiles` is to `:Files`?

junegunn:
Thanks, but I've decided not to add features that I don't actually need/use. I just don't have the mental bandwidth to deal with all the different requirements.

I could make `s:all_files` function public if that's gonna help in your case.

jesseleite:
That's fair.  Making `s:all_files()` public would indeed help a lot!  It's a weird name for what it does though.  Might make more sense if it was made public and renamed to something like `recent_files()`... Thoughts?

jesseleite:
...Also in a [more recent commit](https://github.com/jesseleite/dotfiles/commit/1dc673472ad9225985735c7084fd7c6025840d56), I also had to copy `s:preview()` from [plugin/fzf.vim](https://github.com/junegunn/fzf.vim/blob/master/plugin/fzf.vim#L42-L48).  Would it be worth making `s:preview()` public as well?  I think anybody who is building their own fzf extensions with previews would find that one helpful.

jesseleite:
Submitted PRs, if that helps 🙂

jesseleite:
Thanks again @junegunn for considering my PR's, and merging that `_recent_files()` one specifically ❤️

[...]

PR #1023, opened 2020-05-12
Make `s:p()` public for user fzf extensions?

jesseleite:
Make `s:p()` public for user fzf extensions?  (See #1013 for use case example.)

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -39,7 +39,7 @@ function! s:defs(commands)
   endfor
 endfunction
 
-function! s:p(bang, ...)
+function! fzf#preview(bang, ...)
   let preview_window = get(g:, 'fzf_preview_window', a:bang && &columns >= 80 || &columns >= 120 ? 'right': '')
   if len(preview_window)
     return call('fzf#vim#with_preview', add(copy(a:000), preview_window))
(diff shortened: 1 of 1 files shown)


junegunn:
The function was added quite recently without too much thought and I'm not exactly happy with the way it works because those hard-coded numbers are meaningless to me now as I have configured my fzf to open in a tmux popup window (`let g:fzf_layout = { 'tmux': '-p90%,60%' }`). It needs further consideration. Actually the function is more like a set of some arbitrary rules that are not universally agreed upon so users don't really have to use it.

jesseleite:
>It needs further consideration.

Fair enough.

jesseleite:
>It needs further consideration.

Fair enough.  Whatever you land on though, it would be nice if it was eventually public, so that user extensions can easily piggy back without having to write/copy `g:fzf_preview_window` logic 🙂

junegunn:
Hi, now we no longer have `s:p()` and `fzf#vim#with_preview` will process `g:fzf_preview_window`.

```vim
let g:fzf_preview_window = ['up', 'ctrl-/']
echo fzf#vim#with_preview()
  " {'options': ['--preview-window', 'up', '--preview', '''/Users/jg/.vim/plugged/fzf.vim/bin/preview.sh'' {}', '--bind', 'ctrl-/:toggle-preview']}

let g:fzf_preview_window = []
echo fzf#vim#with_preview()
  " {'options': ['--preview-window', 'hidden', '--preview', '''/Users/jg/.vim/plugged/fzf.vim/bin/preview.sh'' {}']}
```

jesseleite:
Right on, thanks for the info!

PR #1024, opened 2020-05-12
Rename and make `s:all_files` public for user fzf extension?

jesseleite:
Rename and make `s:all_files` public for user fzf extensions? (See #1013 for context and use case example.)


junegunn:
I would name it `fzf#vim#_recent_files` and keep it "semi-private" and undocumented. The nature of the function is quite different from the documented public functions that directly start fzf such as `fzf#vim#files`.

jesseleite:
Done 🤘

PR #1025, opened 2020-05-12
Make `s:get_git_root()` public (or rather semi-private) for user extension?

jesseleite:
I'm working on a few fzf extensions that need to reference the git root, and am having to copy your `s:get_git_root()` function in all of them.  What about making this one public as `_get_git_root()` (or rather semi-private, and obviously undocumented) for user extension?

PS. I know when users like me extend this stuff, it's without warranty 😂, but so many of your functions are script-local, it can be hard to extend at times.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -543,13 +543,13 @@ endfunction
 " GFiles[?]
 " ------------------------------------------------------------------
 
-function! s:get_git_root()
+function! fzf#vim#_get_git_root()
   let root = split(system('git rev-parse --show-toplevel'), '\n')[0]
   return v:shell_error ? '' : root
 endfunction
 
 function! fzf#vim#gitfiles(args, ...)
-  let root = s:get_git_root()
+  let root = fzf#vim#_get_git_root()
   if empty(root)
     return s:warn('Not in git repo')
   endif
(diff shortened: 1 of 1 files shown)


junegunn:
It's a trivial function and I don't think it's necessary to make it public. And you might want to look at Fugitive APIs instead.

Making things public is not an easy decision for names. Once I make such a decision I can't take it back even if it turned out to be a horrible one (which happens all the time) and I have to pay extra care whenever I change the code to maintain backward compatibility and that's no fun. I have to say that I've already pretty much burned out maintaining this project (this was meant to be a mere sample project to demonstrate things you can do with fzf in Vim but look at the number of issues) and I would avoid anything that could add maintenance burden at all costs.

jesseleite:
> this was meant to be a mere sample project to demonstrate things you can do with fzf

Fair enough @junegunn, and I realize this is in your readme as well...

> They are not designed to be flexible or configurable, and there's no guarantee of backward-compatibility.

But then [your readme is also full of extension / customization vimscript examples](https://github.com/junegunn/fzf.vim#advanced-customization), ie)... 

image

And finally...

> I have to say that I've already pretty much burned out maintaining this project [...] and I would avoid anything that could add maintenance burden at all costs.

Burnout sucks.  I apologize for the trouble/burden.  That said, I'm extremely thankful for everything you've done for the vim and cli world ❤️❤️❤️.  Anyway, have a great day!

PR #1026, opened 2020-05-13
[fzf#vim#with_preview] support wsl bash

janlazo:
Close https://github.com/junegunn/fzf.vim/issues/988
Close https://github.com/junegunn/fzf.vim/pull/1009

@ladevieq Please review.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -41,7 +41,6 @@ if s:is_win
   else
     let s:bin.preview = fnamemodify(s:bin.preview, ':8')
   endif
-  let s:bin.preview = 'bash '.escape(s:bin.preview, '\')
 endif
 
 let s:wide = 120
@@ -76,6 +75,11 @@ endfunction
 
 " [[options to wrap], [preview window expression], [toggle-preview keys...]]
 function! fzf#vim#with_preview(...)
+  let bash_path = exepath('bash')
+  if empty(bash_path)
+    throw 'bash is not in PATH'
+  endif
+  let is_wsl_bash = bash_path =~? 'Windows[/\\]system32[/\\]bash.exe$'
   " Default options
   let options = {}
   let window = ''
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -76,6 +75,11 @@ endfunction
 
 " [[options to wrap], [preview window expression], [toggle-preview keys...]]
 function! fzf#vim#with_preview(...)
+  let bash_path = exepath('bash')
+  if empty(bash_path)
+    throw 'bash is not in PATH'
+  endif
+  let is_wsl_bash = bash_path =~? 'Windows[/\\]system32[/\\]bash.exe$'
```
janlazo, line 82: Can wsl bash appear in another directory?

vivlim: There are other commands you can invoke which will result in WSL running, such as `wsl` or distro-specific exes, but AFAIK they normally won't be named bash.exe, so `system32/bash.exe` is what you'll get from calling `bash`. (note: I work for Microsoft, but not on anything related to WSL, so I'm not an authoritative source on this)

As an example of distro-specific exes, I have [VoidWSL](https://github.com/am11/VoidWSL) installed at `c:/wsl/void.exe` and can run commands there such as `c:/wsl/void.exe run path/to/preview.sh /mnt/c/file/to/preview`.

It might be nice to be able to configure the command I want to use if I want to use a specific distro or if I just want to use git bash without rearranging my PATH.

vivlim:
I happened to find your PR while looking into how to fix broken previews on my machine, there are a couple issues I hit:

1. When vim-plug installed fzf.vim, the repo was cloned using my global config `core.autocrlf=true`. This is fine for the plugin itself, because neovim is running in windows, but not okay for the bash script which is running in wsl. WSL detection was added to vim-plug to fix https://github.com/junegunn/vim-plug/issues/821, but that didn't apply in this case because vim-plug had no way of knowing that the repo contained a script which was intended to run in WSL. It ran after I manually fixed the line endings, but I'm not sure how to solve this generally.

2. For me, preview.sh was passed a relative path, not an absolute one-- resulting in WSL not being able to locate the file. I'm not sure why this happened, but since WSL isn't aware of the working directory from windows, it'll always be necessary to pass an absolute path to preview.sh.

janlazo:
1. fzf.vim can add a `.gitattributes` to set the line endings for bash scripts so `core.autocrlf` is no-opt for bash scripts.

2. Latest commit should make fzf.vim use absolute path for the preview script.

junegunn:
@janlazo Thanks. Looks good to me, though as a non-Windows user I can only check if the code is correct.

yum-feng:
I'm having some issues with WSL's not handling fzf's highlighted lines' path properly if there are `\` in it. Though it seems that the bash that comes with git is able to handle it better. **msys** has a newer coreutils but I'm not sure if it makes a difference really.

# WSL

WSL

# msys
gitbash

PR #1038, opened 2020-05-25
fix: Handle absolute windows paths in Grep/Ag/Ripgrep sink process

### LINKED ISSUE #1037: Ripgrep/Grep sink implementation can't handle Windows paths

Ripgrep/Grep sink implementation can't handle Windows paths.

I have implemented HelpRg command that is helpgrep like command but for FZF.

While it works in my WSL environment it fails to open help files in Windows.

```
"" HelpRg command -- like helpgrep but with FZF and ripgrep
let g:helppaths = uniq(sort(split(globpath(&runtimepath, 'doc/', 1), '\n')))
command! -bang -nargs=* HelpRg
            \ call fzf#vim#grep(
            \   'rg --column --line-number --no-heading --color=always --smart-case -g "*.txt" "" '. join(g:helppaths), 1,
            \   {}, <bang>0)

```

The issue is in:

https://github.com/junegunn/fzf.vim/blob/7a655179a43cd431862a2bf5d297d9e55fd3f814/autoload/fzf/vim.vim#L672

After the split of the line `C:/Users/maksim.kim/vimfiles/pack/minpac/opt/fzf/doc/fzf.txt:30:2` instead of "filename" parts[0] has `C`.

vim-fzf-sink-issue

habamax:
Function `ag_to_qf` is used to parse line produced by Grep, Ag or
Ripgrep. Implementation in quesition split it using regexp that doesn't
work for windows absolut paths.

A better approach is to use `matchlist` function.

Fixes #1037

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -669,11 +669,11 @@ endfunction
 " Ag / Rg
 " ------------------------------------------------------------------
 function! s:ag_to_qf(line, has_column)
-  let parts = split(a:line, '[^:]\zs:\ze[^:]')
-  let text = join(parts[(a:has_column ? 3 : 2):], ':')
-  let dict = {'filename': &acd ? fnamemodify(parts[0], ':p') : parts[0], 'lnum': parts[1], 'text': text}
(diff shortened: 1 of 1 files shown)


junegunn:
Merged, thanks!

PR #1044, opened 2020-06-01
Expose some buffer related functions

doronbehar:
Hi!

I really enjoy this plugin. I just wrote a customization to the `:Buffers` command I shared at https://github.com/junegunn/fzf/wiki/Examples-(vim)#delete-buffers-from-buffers and this requires these small changes. I hope it'll be OK with you.

Regards.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -472,7 +472,7 @@ endfunction
 function! s:all_files()
   return fzf#vim#_uniq(map(
     \ filter([expand('%')], 'len(v:val)')
-    \   + filter(map(s:buflisted_sorted(), 'bufname(v:val)'), 'len(v:val)')
+    \   + filter(map(fzf#vim#_buflisted_sorted(), 'bufname(v:val)'), 'len(v:val)')
     \   + filter(copy(v:oldfiles), "filereadable(fnamemodify(v:val, ':p'))"),
     \ 'fnamemodify(v:val, ":~:.")'))
 endfunction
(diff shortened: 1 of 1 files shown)


junegunn:
See my previous comment: https://github.com/junegunn/fzf.vim/pull/1024#issuecomment-627323510

Making things public means that I'll have to pay extra care not to break backward compatibility from then on, and I don't want to do it and I want to make it clear by naming these functions with `_` prefix.

doronbehar:
Done :guitar:.

autoload/fzf/vim.vim
```diff
@@ -640,15 +640,15 @@ function! s:sort_buffers(...)
   return b1 < b2 ? 1 : -1
 endfunction
 
-function! s:buflisted_sorted()
+function! fzf#vim#_buflisted_sorted()
```
junegunn, line 643: There is still `s:buflisted_sorted` in the code.

doronbehar: :+1: I think I catched them all (with git grep).

yanzhang0219:
Hi @doronbehar 

I encounter an error when I run the customized version of `:Buffers` command and try to delete a buffer using `D`. I attached the screenshot below. Thank you.

image

doronbehar:
Yea I know, another person has contacted me personally to note this fault. However, from my experience these errors are not consistent - they don't appear always and they are not always reproducible, at least for me.

The problem is, that when `D` is pressed, we want to change something both in "the parent" Nvim session and in the list of open buffers in fzf "the child". This makes it hard to decide whether `D` should be mapped to something by an `autocmd` (By Vim) - which will make it easy to delete buffers with it, or to use `fzf`'s `--bind`. In the current implementation, the latter choice was made.

I'm currently unavailable to solve this little quirk, though it's in my TODO now. I will find the time probably at the end of July, when my semester will over.

P.S, sorry `@junegunn` for abusing your repo to discuss this.

yanzhang0219:
Okay. I think it is better to give a tag/mark like "bug" to this customized command in the wiki page so that other users could recognize it and wait for it being solved. Thanks.

doronbehar:
I've removed the wiki section as I've reached the conclusion it'll be impossible to implement this idea without the feature described in https://github.com/junegunn/fzf/issues/2019 .

PR #1067, opened 2020-07-17
Fix :Maps when :verbose map does not have a 2nd line with comments

antoinemadec:
We were only adding elements in the list outside of the for loop.


antoinemadec:
@junegunn , did you get a chance to look at this?
This is a very minimal change with great benefits 🤓

junegunn:
I don't have any maps with newlines. How can I test this? And can you give an example of `:verbose map` output that is causing the issue?

antoinemadec:
@junegunn , here is the minimal **vimrc** to reproduce this:
```vim
call plug#begin('./plugins')
Plug 'neoclide/coc.nvim', {'branch': 'release'}
Plug 'junegunn/fzf', {'dir': '~/.fzf','do': './install --all'}
Plug 'junegunn/fzf.vim'
call plug#end()

set nocompatible
```

The issue is that this command does not have the same output on **vim** and **neovim**:
```vim
:verbose imap
```

Try to look for **BS** after running ```call fzf#vim#maps('i')```, in **vim** you will find it, not in **neovim**.
Note that my PR is not good as it doubles a lot of the mapping (with and without the file/line number).

antoinemadec:
My PR is now fixing the issue properly, no more duplicate.
The actually issue was the lack of 2nd line with comments in neovim when running ```:verbose imap```.

This happens when creating mappings with **api.vim**.

junegunn:
Okay, thanks!

Issue #1052, opened 2020-06-15
Windows subsystem support breaks older vim


vyasr:
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

The changes made in #1026 seem to have broken support for old versions of vim. Specifically, the use of the `exepath` command in the `fzf#vim#with_preview` fails because the command does not appear to exist in vim 7.4 (based on both trying to execute the command and parsing the help page for eval.txt, where I `exepath` defined in vim 8.2). I don't know if there is still interest in this plugin supporting such an old version of vim, but I unfortunately do some work on systems where old versions are not uncommon. I should note that while executing `:Files` does cause fzf to print an error message, it still seems to generally work as expected afterwards (it generates a list of files to select and open).

For what it's worth, I think this error could be fixed by simply replacing `exepath('bash')` with `system('which bash')` (I've made this change locally), but I don't know if that possibly causes other problems. There may be a more vim-specific solution, but I'm not particularly familiar with the functions vim-script has available for this purpose.

janlazo:
`exepath` was added in Vim v7.4.0235. Which Vim version are you running?

vyasr:
That's odd, I ran into this issue running vim 7.4.1099. Here's the output of `vim --version` if that's helpful:
```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Oct 30 2018 19:56:57)
Included patches: 1-160, 399, 402-403, 1099
Modified by <bugzilla@redhat.com>
Compiled by <bugzilla@redhat.com>
Huge version without GUI.  Features included (+) or not (-):
+acl             +farsi           +mouse_netterm   +syntax
+arabic          +file_in_path    +mouse_sgr       +tag_binary
+autocmd         +find_in_path    -mouse_sysmouse  +tag_old_static
-balloon_eval    +float           +mouse_urxvt     -tag_any_white
-browse          +folding         +mouse_xterm     -tcl
++builtin_terms  -footer          +multi_byte      +terminfo
+byte_offset     +fork()          +multi_lang      +termresponse
+cindent         +gettext         -mzscheme        +textobjects
-clientserver    -hangul_input    +netbeans_intg   +title
-clipboard       +iconv           +path_extra      -toolbar
[...]

PR #1121, opened 2020-09-11
[fzf#vim#with_preview] Detect WSL bash on Windows only

janlazo:
Close https://github.com/junegunn/fzf.vim/issues/1052

@vyasr Can you test?


vyasr:
Yes, I can confirm that this works. Thanks for the patch!

junegunn:
Merged, thanks! 🙏

PR #1142, opened 2020-10-20
Concat string with "."

sicher:
The ".." Lua operator has sneaked in here, causing invalid expression errors... :)

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -72,7 +72,7 @@ function! s:check_requirements()
     throw "fzf#exec function not found. You need to upgrade Vim plugin from the main fzf repository ('junegunn/fzf')"
   endif
   let exec = fzf#exec()
-  let fzf_version = matchstr(systemlist(exec .. ' --version')[0], '[0-9.]*')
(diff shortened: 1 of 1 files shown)


junegunn:
Vim version?

```
:help expr-..

expr5 and expr6						*expr5* *expr6*
---------------
expr6 + expr6   Number addition, |List| or |Blob| concatenation	*expr-+*
expr6 - expr6   Number subtraction				*expr--*
expr6 . expr6   String concatenation				*expr-.*
expr6 .. expr6  String concatenation				*expr-..*
```

junegunn:
Okay, looks like `..` operator is not available on Vim 7.4. While I don't guarantee that this plugin works on older versions of Vim, I'm going to merge this as it's no big deal. Thanks.

sicher:
> Vim version?

I am on 8.1.1401 (official Debian 10 stable package) which should have the operator but doesn't (weird!).

It seems the operator was added to vim in patch 8.1.1114 (see https://github.com/vim/vim/commit/0f248b006c2574abc00c9aa7886d8f33620eb822)

PR #1143, opened 2020-10-21
Fix regex to extract fzf_version

dsewnr:
I got an error while version check.

`E605: Exception not caught: You need to upgrade fzf. Found:  (/Users/peter_br_lu/.fzf/bin/fzf). Required: 0.23.0 or above.`

My fzf version from cli:
```
⟩ /Users/peter_br_lu/.fzf/bin/fzf --version
0.23.1 (fc7630a)
```
and try to fix....

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -72,7 +72,7 @@ function! s:check_requirements()
     throw "fzf#exec function not found. You need to upgrade Vim plugin from the main fzf repository ('junegunn/fzf')"
   endif
   let exec = fzf#exec()
-  let fzf_version = matchstr(systemlist(exec .. ' --version')[0], '[0-9.]*')
+  let fzf_version = matchstr(systemlist(exec .. ' --version')[0], '[0-9.]\+')
 
   if s:version_requirement(fzf_version, s:min_version)
     let s:checked = 1


junegunn:
I'm not sure what the problem is

```vim
:echo matchstr('0.23.1 (fc7630a)', '[0-9.]*')
0.23.1
```

dsewnr:
I try to change regex to `.*[0-9.]*` and get an unexpected character before version string
`E605: Exception not caught: You need to upgrade fzf. Found: ^[[m^O0.23.1 (fc7630a) (/Users/peter_br_lu/.fzf/bin/fzf). Required: 0.23.0 or above.`

junegunn:
Well, then your shell configuration is broken and you have to fix it.

- What is your `:echo &shell`?
- You have the same problem when you run `:echo string(system('echo -n foo'))`, right?

dsewnr:
`:echo &shell` => /usr/local/bin/fish

`echo string(system('echo -n foo'))` => '^[[m^Ofoo'

OK, I'll try to figure out what the problem from my shell.

dsewnr:
I found the problem is caused by my Neovim (NVIM v0.4.4). 
Everything works fine from my Vim 8.2 (VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Oct 15 2020 23:34:31)).

junegunn:
This may give you some hint. https://github.com/junegunn/fzf/issues/1415

dsewnr:
Thanks for the hint.
I found the plugin `oh-my-fish/plugin-vi-mode` is the root cause.
I've just removed it and enjoy fzf. 👍 

Thanks again.

PR #1159, opened 2020-11-02
Incorporate batcat name

DZ84:
Seemed to me this is neater then changing it manually, or having to create a symlink, which isn't necessarily possible when user has installed the alternative bat package.

Additionally/alternatively a vimrc option could also be created to define path/name of executable of bat.


junegunn:
Merged, thanks.

PR #1160, opened 2020-11-03
[fzf#vim#gitfiles] Add arg for changing root

imran-iq:
Hi,

First of all thank you for creating `fzf` its completely changed the way I use vim and the command line.

In my current role, I am stuck working on a mono repo and `:GFiles` ends up grabbing all files from all the projects in the mono repo (which is not that ideal).

This change adds a new command `GFilesCurrentDir` which works essentially the same as `:GFiles` but using the current directory as the `dir` for the `fzf` call instead of the git project root. The end results (for mono repos) is that `:GFilesCurrentDir` in a project directory works the same as `:GFiles` in a regular git repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,32 +59,34 @@ so you can omit it if you use a plugin manager that doesn't support hooks.
 Commands
 --------
 
-| Command           | List                                                                    |
-| ---               | ---                                                                     |
-| `:Files [PATH]`   | Files (runs `$FZF_DEFAULT_COMMAND` if defined)                          |
-| `:GFiles [OPTS]`  | Git files (`git ls-files`)                                              |
-| `:GFiles?`        | Git files (`git status`)                                                |
-| `:Buffers`        | Open buffers                                                            |
-| `:Colors`         | Color schemes                                                           |
-| `:Ag [PATTERN]`   | [ag][ag] search result (`ALT-A` to select all, `ALT-D` to deselect all) |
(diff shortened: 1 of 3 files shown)


junegunn:
Hi, @imran-iq 

Thanks for the contribution and the support.

I don't think we need to provide a separate command for that purpose. You can define your own variation of the command using the autoload function. In this case, you want to pass `dir` option to the spec dictionary.

```vim
command! -bang -nargs=? GFilesCwd
  \ call fzf#vim#gitfiles(<q-args>, fzf#vim#with_preview(<q-args> == '?' ? { 'dir': getcwd(), 'placeholder': '' } : { 'dir': getcwd() }), <bang>0)
```

The empty `placeholder` part for `:GF?` can be hard to understand as it's undocumented. The purpose of it is not to add `--preview` option to the spec so that the special preview command for `git status` (one using `delta`) is not overridden.

```vim
" See how the function works
:echo fzf#vim#with_preview({})
:echo fzf#vim#with_preview({'dir': getcwd()})
:echo fzf#vim#with_preview({'dir': getcwd(), 'placeholder': ''})
```

tmccombs:
@junegunn I just tried that `GFilesCwd` command, and it doesn't seem to work, it still uses the git root.

tmccombs:
Could that `GFilesCwd` command be added tot he plugin?

PR #1162, opened 2020-11-03
Disabled automatic column option for Ag.

DZ84:
The build in grep function doesn't automatically show the column number, and therefore the Rg and GGrep examples in the README don't either. But Ag does. This branch removes this default option.

Not sure if there are any backwards compatibility problems, but don't see why it should and for me it works fine.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -767,7 +767,7 @@ function! fzf#vim#ag_raw(command_suffix, ...)
   if !executable('ag')
     return s:warn('ag is not found')
   endif
-  return call('fzf#vim#grep', extend(['ag --nogroup --column --color '.a:command_suffix, 1], a:000))
+  return call('fzf#vim#grep', extend(['ag --nogroup --color '.a:command_suffix, 1], a:000))
 endfunction
 
 " command (string), has_column (0/1), [options (dict)], [fullscreen (0/1)]


junegunn:
The column number will bring your cursor to the exact location of the match, so it's nice to have. Maybe you run the command with an empty query string?

DZ84:
Yeah, that is exactly what I did, which is why I got all the 1’s, which I then wanted to get rid of. I agree, that is nice to have.

However, I would still like to make the case for uniformity: if, when you set up Rg, you have to enable the columns option, why not for Ag? Also, I don’t see a way to disable it when it’s already specified. Rg has a --no-column flag, Ag doesn’t.

junegunn:
> therefore the Rg and GGrep examples in the README don't either.

Not true. Rg example sets the second argument to `fzf#vim#grep` to `1`. The example also explains why the argument is not set for GGrep.

> > The second argument to `fzf#vim#grep` is 0 (false), because git grep does not print column numbers.

> when you set up Rg, you have to enable the columns option

Not true. It's up to you to pass the right command and flag to `fzf#vim#grep`. If your command prints column numbers, pass 1 as the second argument. If you use a command that doesn't print column numbers, pass 0.

> I don’t see a way to disable it when it’s already specified.

You can define your own Ag command using `fzf#vim#grep` function just like the examples in the README page.

DZ84:
> Not true. Rg example uses a command with --column and sets the second argument to fzf#vim#grep to 1. The page also explains why the argument is not set for GGrep.

The --column with second argument 1 was exactly what I meant with 'enabling'.

> > > The second argument to fzf#vim#grep is 0 (false), because git grep does not print column numbers.

I hadn't catched that GGrep information in the README, but I guess it's outdated: git grep can now forward column numbers. Which is why I opened pr [#1167](https://github.com/junegunn/fzf.vim/pull/1167).

> You can define your own Ag command using fzf#vim#grep function just like the examples in the README page.

Makes sense.

PR #1167, opened 2020-11-05
Adjusted the README and doc to adjust to new situation git grep.

DZ84:
Adjusted README and fzf-vim.txt to reflect up to date situation git grep functionality.


DZ84:
Is this still under consideration? Seems to me like a fairly straightforward update of the readme content.

PR #1175, opened 2020-11-10
Allow bat --color to be user configurable

connorlay:
This PR allows the `--color` option for `bat` to be defined by the user, so it can be disabled. `bat` supports a [configuration file](https://github.com/sharkdp/bat#configuration-file) that is automatically picked up if present. The default behavior of `bat` is to display text with colors, so this would not break backwards compatibility by removing the explicit `--color=always` flag.

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -47,8 +47,7 @@ elif command -v bat > /dev/null; then
 fi
 
 if [ -z "$FZF_PREVIEW_COMMAND" ] && [ "${BATNAME:+x}" ]; then
-  ${BATNAME} --style="${BAT_STYLE:-numbers}" --color=always --pager=never \
-      --highlight-line=$CENTER "$FILE"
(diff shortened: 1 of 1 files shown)


junegunn:
> The default behavior of bat is to display text with colors

Only if its standard output is a TTY device, which is not the case here.

junegunn:
```
       --color <when>

              Specify when to use colored output. The automatic mode only enables colors if an interactive terminal is detected. Possible values: *auto*, never, always.
```

connorlay:
Oh good find, what do you think about making this configurable some other way then? I am using a terminal and vim color scheme that visually is more appealing when bat is configured to display without colors. I tried using the configuration file to set `--color=never`, but that is being overwritten by the CLI option.

junegunn:
Well, then there's really no point in using `bat` in the first place, right? Maybe you could choose a `BAT_THEME` that looks better in your setup.

connorlay:
That's true, I had not thought of dropping `bat` and using `cat` instead. I'll go ahead and do that instead. Thanks!

PR #1177, opened 2020-11-11
Fix :Tags returning no results when the current buffer is a Neovim terminal

jeromedalbert:
## Problem

`tagfiles()` returns paths that are prefixed with `term://` when the current buffer is a Neovim terminal. But those are not valid paths, so ctags ignores them, which means that `:Tags` / `fzf#vim#tags` returns no results.

## Solution

Fix the problem by removing the `term://` prefix from these paths.


jeromedalbert:
This PR is stale, closing it.

PR #1189, opened 2020-11-27
Fix file preview on Windows

brainplot:
On Windows, whenever I tried to use any of the commands to pick files, in the preview I would get an error saying that the target file was not found, but *only* when the target file was in a directory. If this sounds confusing, I'm providing screenshots below:

outer
_(file located in the project's root)_

inner
_(file located in a subdirectory)_

I figured somehow a double-backslash was being used as a path separator, on Windows. I looked at the `preview.sh` script and replaced all pairs of backslashes in `$FILE` with a single forward slash, which Windows also accepts.

I repointed the plugin to my local repository and it seems to be working.
Unfortunately I don't have a Unix system at hand at the moment to test it on but _it shouldn't™_ break anything there.

Thanks for your work on `fzf.vim`! I literally wouldn't use Vim as my code editor without it 🙂

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -23,6 +23,7 @@ fi
 CENTER=${CENTER/[^0-9]*/}
 
 FILE="${FILE/#\~\//$HOME/}"
+FILE="${FILE//\\\\/\/}"
 if [ ! -r "$FILE" ]; then
   echo "File not found ${FILE}"
   exit 1


junegunn:
> I figured somehow a double-backslash was being used as a path separator, on Windows.

fzf escapes backslashes when building the preview command, so I guess that's why you're getting double backslashes.

https://github.com/junegunn/fzf/blob/298342677107f19f2ef7035afcf5302fdf26aa51/src/terminal.go#L1457-L1464

@janlazo Hi, any idea how we can address this?

janlazo:
Most programs on Windows, like powershell, must escape double-quote and backslash characters with a backslash for its arguments. cmd.exe has its own escaping rules so it does not use backslash to escape its special characters.  vim-plug does not escape backslashes for cmd.exe arguments. If I had time, I'd update fzf to use the same escaping code as vim-plug in the Vimscript plugin and the golang code and then add additional escaping for Vim's `system()` and `:!`.

It can be breaking change for preview commands that run nested executables (ie. `sh -c "sh -c "`) or run mingw/msys2/WSL programs like bash. A test case could be executing the selected file, with unescaped filepath, using powershell or WSL bash.

ayushxx7 (commented):
I was not seeing the preview window (Windows 10), so after looking at the issues I first removed the blank lines. 
After which I faced this issue, which was fixed by adding this line mentioned in the commit.
```
FILE="${FILE//\\\\/\/}"
```
However, I cannot get syntax highlight to work. 
From what I have tested, vim/fzf/preview.sh is not able to pick `bat` but I can access it directly in the terminal. 

Any suggestions?

brainplot:
On some distros, such as  Ubuntu, `bat` is installed as `batcat` due to a conflict with another package. Since the `preview.sh` script runs in Bash (because of the `#!/usr/bin/env bash` shebang at the top) it might be that you're not using Bash as your shell in your terminal but something else (maybe zsh if you are on a Mac?).
If you've set some kind of alias for `bat`, it would not be picked up by the script when it runs.

That would be my guess, at least.

ayushxx7:
> On some distros, such as Ubuntu, `bat` is installed as `batcat` due to a conflict with another package. Since the `preview.sh` script runs in Bash (because of the `#!/usr/bin/env bash` shebang at the top) it might be that you're not using Bash as your shell in your terminal but something else (maybe zsh if you are on a Mac?).
> If you've set some kind of alias for `bat`, it would not be picked up by the script when it runs.
> 
> That would be my guess, at least.

I am using Windows10. If I type `bat` in my terminal, it works correctly. I have a raised an issue referencing [this PR](https://github.com/junegunn/fzf.vim/issues/1207).

brainplot:
Does `bat` (without `.exe`) work in Git Bash?

ayushxx7:
> Does `bat` (without `.exe`) work in Git Bash?

TL;DR: Had to install `bat` in WSL using `sudo apt install bat`. (Screenshot attached at the end)

So, as it turns out, when I type in `bash` in Command Prompt, it would load the WSL (Windows Subsystem for Linux). Then I checked if I could open `bat` there, as you suggested, but it failed. I then installed `bat` using: 

```sudo apt install bacula-console-qt``` -- DO NOT DO THIS. I blindly copied from the command prompt thinking that was the full-form of bat...

Then I ended up with this error:
```
bat: error while loading shared libraries: libQt5Core.so.5: cannot open shared object file: No such file or directory
```

Found a workaround, by running this command:
```
sudo strip --remove-section=.note.ABI-tag /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
```
Only to end up with this error:
```
qt.qpa.xcb: could not connect to display
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
```
when running `bat` from WSL, but the fzf/rg preview now has syntax highlighting

Finally, I installed `bat` again using 
```sudo apt install bat```
This installed `bat` as `batcat`, and `preview.sh` was able to pick it up. Now, I have syntax highlighting!
image

Many thanks for your help @brainplot!

Grueslayer:
This is only half of the problem, please see my PR #1215 solves your problem, too.

Issue #1031, opened 2020-05-16
Not possible to jump to existing open tab using :Lines


dpacbach:
Hi,

After much searching and trying various solutions posted in forums, I have been unable to get the `:Lines` command to allow me to jump to an existing tab when the chosen file is already open in another tab.  Is there any way to do this?  I tried setting `g:fzf_buffers_jump = 1` but that does not seem to help.  I also tried various things with `g:fzf_action` but couldn't make that work given the implementation of `s:line_handler`.  Can someone suggest a way?  Here is the current implementation of the `s:line_handler`:

```vim
function! s:line_handler(lines)
  if len(a:lines) < 2
    return
  endif
  normal! m'
  let cmd = s:action_for(a:lines[0])
  if !empty(cmd) && stridx('edit', cmd) < 0
    execute 'silent' cmd
  endif

  let keys = split(a:lines[1], '\t')
  execute 'buffer' keys[0]
  execute keys[2]
  normal! ^zvzz
endfunction
```

but as can be seen, the `execute 'silent' cmd`, while it will execute a command that I put into `g:fzf_action`, does not allow passing the name of the file selected, so I can't write a custom function to be executed there.  Can someone help?

- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

przepompownia:
With `let g:fzf_action['enter'] = 'tab drop'`I have similar problem with, for example `:Buffers` which use `s:bufopen` with the same `execute 'silent' cmd`. I get `Vim(drop):E471: Argument required: silent tab drop`. 

Results from `fzf#vim#files('.')` are correctly opened because of no sink function similar to mentioned above.

iiey:
It would be awesome if someone could fix it for `:Buffers` too.

Problem can reproduce with minimal config:
```
let g:fzf_action = {
    \ 'ctrl-t': 'tab drop',
    \ 'ctrl-x': 'split',
    \ 'ctrl-v': 'vsplit' }
```
Then try `ctrl-t` on any buffer after executing `:Buffers` causes error:

```
Error detected while processing function fzf#vim#buffers[3]..<SNR>143_fzf[18]..fzf#run[63]..<SNR>21_callback:
Vim(drop):E471: Argument required: silent tab drop
```

[...]

PR #1192, opened 2020-11-30
Fix 'tab drop' action for :Buffers

przepompownia:
Related to #1031

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -674,10 +674,20 @@ function! s:bufopen(lines)
     endif
   endif
   let cmd = s:action_for(a:lines[0])
-  if !empty(cmd)
-    execute 'silent' cmd
+
+  call s:open_buffer_with_action(b, cmd)
+endfunction
+
+function! s:open_buffer_with_action(buffer, action) abort
+  if 'tab drop' ==# a:action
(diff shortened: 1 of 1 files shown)


iiey:
Could contributor preview for merging this pull request?
Last time I check with master (e34f6c129d39b90db44df1107c8b7dfacfd18946), it's still working correctly.

hejops:
Is this ever going to get merged?

iiey:
Is there an explanation why this is closed without merging?

przepompownia:
@iiey it was done automatically by deletion the whole repository but I see that the change is still visible here. About a year ago I replaced FZF by Telescope (written in Lua, Neovim only) with https://github.com/nvim-telescope/telescope-fzf-native.nvim. Most outstanding issues for fzf(.vim), especially numerous crashes hard to reproduce (because of randomness) no longer occur for me. Telescope can work without using built-in terminal. I still use FZF in bash, but not for vim. 

> This repository is a bundle of fzf-based commands and mappings extracted from my [.vimrc](https://github.com/junegunn/dotfiles/blob/master/vimrc) to address such needs. They are not designed to be flexible or configurable, and there's no guarantee of backward-compatibility.

Maybe we should treat this repo as the README says (a set of examples to modify) and apply requested changes in our own repositories.

Issue #595, opened 2018-03-13
Jump list is not used with fzf#vim#grep


RobertBuhren:
- Category
    - [ ] Question
    - [ ] Bug
    - [x] Suggestion
- OS
    - [x] Linux
    - [ ] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [x] Vim
    - [x] Neovim

When using the builtin "Ag" cmd or the "Rg" command from the help file, the jump list is not used. So when I'm at line containing "foo" and I to e.g. "Ag bar", I can't use "CTRL-O" to go back to "foo".

PR #1195, opened 2020-12-03
set jump if the target is in the current file

magras:
Adding a record to the jump list allows to go back to the cursor position before jump with CTRL-O.


magras:
This PR resolves #595.

Probably my description of the problem was not clear enough, so may be a link to the issue will clarify why I've made this change.

junegunn:
Merged, thanks!

Issue #796, opened 2019-04-11
Snippet expansion and word under cursor


aldantas:
- Category
    - [ ] Question
    - [ ] Bug
    - [x] Suggestion
- OS
    - [x] Linux
    - [ ] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [x] Vim
    - [ ] Neovim

When the Snippet command is called, the word under cursor is automatically used as initial pattern for the search. The problem is that, once a matching is selected, the snippet keyword  is inserted after the current word and hence the snippet expansion fails.

For example, if the word under cursor is  'def', one possible suggestion could be 'defm'. Upon selecting the later, the resulting text becomes 'defdefm', which is not a valid snippet keyword.

So perhaps there could be a way to prevent this, either by deleting the current word under cursor before inserting the selected keyword, or by completing the keyword instead of inserting it whole. The former solution is simpler, just replace

```
function! s:inject_snippet(line)
  let snip = split(a:line, "\t")[0]
  execute 'normal! a'.s:strip(snip)."\<c-r>=UltiSnips#ExpandSnippet()\<cr>"
endfunction
```

to

```
function! s:inject_snippet(line)
  let snip = split(a:line, "\t")[0]
  execute 'normal! diwi'.s:strip(snip)."\<c-r>=UltiSnips#ExpandSnippet()\<cr>"
endfunction
```

maamo888:
Will this functionality change not be reflected?

I also prefer this behavior for `:Snippets`.

I use it in the insert mode, but now I use it with caution. :(

```
inoremap <c-f><c-u> <c-\><c-o>:Snippets<CR>
```

@junegunn @aldantas

josefson:
Came here looking for this as well, i'd really like for this to work out.
Right now i've been using this which works really well, but i dont have anyway of filtering the list down AND i gotta select with numbers.
```inoremap <silent> <C-j> <C-r>=utils#LoadUltiSnipsAndExpand()<CR>```

liskin:
[...]

PR #1196, opened 2020-12-04
[Snippets] Fix entering insert mode and handling of word under cursor

liskin:
Two fixes for #796

[Snippets] Fix UltiSnips not being able to enter insert mode

UltiSnips snippets have so called tabstops or placeholders that are to be filled in by the user, so after expanding a snippet the cursor is moved to the first such tabstop and vim enters insert mode (or, if there's a default value, select mode).

But this doesn't work with fzf's :Snippets, because UltiSnips uses :startinsert which doesn't work from :normal. Fix this by calling UltiSnips#ExpandSnippet directly (unfortunately needs a virtualedit hack to position the cursor after the snippet name).

---

[Snippets] Fix when there is word under cursor

UltiSnips#SnippetsInCurrentScope uses the word under cursor, if any, to filter candidate snippets. To make it usable, however, fzf's :Snippets must replace the word instead of appending to it, otherwise the subsequent expansion fails.

Also, use 'i' instead of 'a', to make mappings such as

	inoremap <silent> <C-_> <C-\><C-O>:Snippets<CR>

feel natural. This will break workflow for people who used to manually exit insert mode in the middle of a line and then invoke :Snippets, but I'd hazard a guess there aren't many of them as this workflow was already somewhat broken by the first bug.

---

Fixes: https://github.com/junegunn/fzf.vim/issues/796

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -943,8 +943,9 @@ endfunction
 function! s:inject_snippet(line)
   let ve = &ve
   set ve=onemore
+  let del = col('.') > 1 && getline('.')[0 : col('.')-2] =~ "\\S$" ? "\<c-w>" : ""
   let snip = split(a:line, "\t")[0]
-  execute 'normal! a'.s:strip(snip)
+  execute 'normal! i'.del.s:strip(snip)
   execute 'normal! l'
   call UltiSnips#ExpandSnippet()
   let &ve = ve
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -941,8 +941,13 @@ endfunction
 " Snippets (UltiSnips)
 " ------------------------------------------------------------------
 function! s:inject_snippet(line)
+  let ve = &ve
+  set ve=onemore
   let snip = split(a:line, "\t")[0]
-  execute 'normal! a'.s:strip(snip)."\<c-r>=UltiSnips#ExpandSnippet()\<cr>"
+  execute 'normal! a'.s:strip(snip)
+  execute 'normal! l'
+  call UltiSnips#ExpandSnippet()
+  let &ve = ve
 endfunction
 
 function! fzf#vim#snippets(...)


junegunn:
Thanks for the patch, but I haven't used UltiSnips for a long time.

@Akuseru1 Does this fix your problem?

liskin:
I intend to discuss this with the UltiSnips folks as it seems I'm still not seeing the whole picture.
I'd welcome your opinion on how safe it is to use startinsert/feedkeys as I have very little experience with those two. Either way, please refrain from merging this just yet, it's likely I'll be pushing improvements. :-)

liskin:
In the meantime I discovered a few things:

1. UltiSnips devs no longer remember the details of why they sometimes startinsert and other times feedkeys("i"), they referred me to try to refactor it and then try the test suite, which I added to my todo list and never got back to :-/
2. vim 8.2.1978 added `<Cmd>` that can be used instead of `<C-\><C-O>` in insert mode bindings without the drawbacks of `mode()` giving misleading answers, so I can leave the UltiSnips code as it is and not worry about it ever again
3. there was an occasional glitch after snippet expansions caused by not invoking the CursorMovedI autocmd hook after inserting the selected snippet name in fzf, which I fixed now

I am not a heavy user of UltiSnips these days (not writing much new code in languages that would benefit from it) so there may still be some issues. I'll ping the folks in #796 to test this.

aldantas:
Hi, thank you for the PR. I haven't been using fzf for searching snippets much lately, but I tested it and it fixes the problem of expansion. Also, the placeholders are working as they should now. I didn't understand what the problem with `<C-\><C-O>` was. For me, it worked with both  `<C-\><C-O>` and `<Cmd>` (and also witrh `<Esc>`).

However, I noticed that it was inserting the last character from the pre-expansion word after the last placeholder. For example, in python, when typing the word `if`, triggering the search and selecting the snippet `ife`, it expands to:
```
if condition:
    pass
else:
    passf
```
And the `f` at the end still remains even after finishing editing the placeholders.

I was able to fix that by changing your code at line:
`let del = empty(matchstr(getline('.'), '\%' . (col('.') - 1) . 'c\S')) ? "" : "\<c-w>"`
to:
`let del = empty(matchstr(getline('.'), '\%' . (col('.')) . 'c\S')) ? "" : "\<c-w>"`

And at line:
`execute 'normal! i'.del.s:strip(snip)`
to:
`execute 'normal! a'.del.s:strip(snip)`

But I don't know if that breaks something else.

liskin:
> I didn't understand what the problem with `<C-\><C-O>` was. For me, it worked with both `<C-\><C-O>` and `<Cmd>` (and also witrh `<Esc>`).

Snippets with placeholders with some default text should not work in this PR with `<C-\><C-O>` (some garbage text appears instead). In the original code, such snippets probably did work, but instead snippets without default placeholder which are supposed to start insert mode didn't start the insert mode properly.

> However, I noticed that it was inserting the last character from the pre-expansion word after the last placeholder. For example, in python, when typing the word `if`, triggering the search and selecting the snippet `ife`, it expands to:
> […]

This is weird. That's exactly what the `<C-\>` in `<C-\><C-O>` is meant to prevent, and `<Cmd>` shouldn't be moving the cursor either. So this should only happen if you `<Esc>` to normal mode and then `:Snippets`. Can you confirm that's what you're seeing or is this happening with `<C-\><C-O>`/`<Cmd>` as well?

I'm not really sure what to do about `:Snippets` in normal mode though. Should we assume the user wants the cursor one column to the right and move it even before `UltiSnips#SnippetsInCurrentScope` is consuled? Or maybe move the cursor to the end of the word?

> I was able to fix that by changing your code at line:
> `let del = empty(matchstr(getline('.'), '\%' . (col('.') - 1) . 'c\S')) ? "" : "\<c-w>"`
> to:
> `let del = empty(matchstr(getline('.'), '\%' . (col('.')) . 'c\S')) ? "" : "\<c-w>"`
> 
> And at line:
> `execute 'normal! i'.del.s:strip(snip)`
> to:
> `execute 'normal! a'.del.s:strip(snip)`
> 
> But I don't know if that breaks something else.

This definitely breaks insert mode mappings for me.

PR #1203, opened 2020-12-12
Add Tselect command to filter matching tags

lakshayg:
This commit adds a command roughly equivalent to :tselect.

The use can now call Tselect with an optional query. If a query
is specified, all matches for that query are looked up and loaded
for filtering into an fzf window. If no query is specified, the top
tag on the tagstack is used (just like tselect)

This commit borrows some code from https://github.com/zackhsi/fzf-tags

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -943,6 +943,49 @@ function! fzf#vim#tags(query, ...)
 endfunction
 
 " ------------------------------------------------------------------
+" Tselect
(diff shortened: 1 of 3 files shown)


faisal-shah:
Any intention to entertain this PR? Are there any changes you would like to see?

lakshayg:
I am closing this PR in favor of https://github.com/junegunn/fzf.vim/pull/1281 which does not conflict with master

PR #1215, opened 2021-01-03
Preview support for MS Windows (WSL and Git bash) - Fixes #1212

### LINKED ISSUE #1212: Preview does not work on MS Win (neither WSL or Git Bash)
- [x] I have fzf 0.23.0 or above

Preview will not run on MS Win systems (maybe Cygwin, couldn't test).

Let's have a look at the situation:

- VIM is running native on MSWin using path with backslash (and drive-letters and/or UNC path)
- FZF is running native on MSWin using same environment
- RG is running native on MSWin using same environment
- All others (like MRU lists) follow these directions

If now FZF calls the `preview.sh` script (by any bash.exe it founds) the argument is a *MS Win* path.

Let's assume we only found the WSL bash (where some exceptions are programmed into the fzf.vim code) it can only use files which are in the local directory. All other files contain the wrong path delimiter (or additional drive / UNC server names).

Only if we are on a bash (say the bash environment coming with the native GIT) that will work.

On WSL we need to *convert* the local MS Win path to a WSL path. Therefore exists a command called `wslpath` you can give a Windows path and receive a Linux path for the current WSL instance.

But there is a second special behaviour: VIM simulates the `~` known from un*x like systems as home directory for the current user. This is accomplished by checking against environment variables:

- `$HOME` (which is normally *not* set on a MS Win system
- `$HOMEDRIVE$HOMEPATH` (which default is the home of the current loggon user)

(It seems VIM does not check $HOMESHARE and does not use $PROFILEPATH)

So if you are using `History` you'll see p.e. `~\_vimrc`.

The WSL bash is a complete linux system which has its own `HOME` which has nothing in common with the windows users home (above).

So we need to tell the `preview.sh` what to do with that `~` in the path, too. The WSL has normally no access to MS Win environment vairiable but you can *forward* those. 

The *MS Win* environment variable (it's a list in un*x style) is to be set:

```
WSLENV=HOMEDRIVE/u:HOMEPATH/u:
```

Here we have two possible solutions for this:

1. The user sets those variables in the "System Enviromnent Variables" dialog from MS Win (WSL always has access to these variables)
2. If fzf.vim detects the WSL bash, it will check the environment  variable WSLENV for HOMEDRIVE and HOMEPATH content. if it does not exists, it adds those like above. The WSL bash has only access to it when called from VIM.

This a prove-of-concept running with WSL and Git bash:

```diff
diff --git a/bin/preview.sh b/bin/preview.sh
index 08a8c07..06587d2 100755
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -22,6 +22,18 @@ if [[ -n "$CENTER" && ! "$CENTER" =~ ^[0-9] ]]; then
 fi
 CENTER=${CENTER/[^0-9]*/}
 
+#MS Win support
+if [ -z "$MSWINHOME" ]; then
+  MSWINHOME="$HOMEDRIVE$HOMEPATH"
+fi
+if grep -qEi "(Microsoft|WSL)" /proc/version &> /dev/null ; then
+  MSWINHOME="${MSWINHOME//\\/\\\\}"
+  FILE="${FILE/#\~\\/$MSWINHOME\\}"
+  FILE=`wslpath -u "$FILE"`
+elif [ ! -z "$MSWINHOME" ]; then
+  FILE="${FILE/#\~\\/$MSWINHOME\\}"
+fi
+
 FILE="${FILE/#\~\//$HOME/}"
 if [ ! -r "$FILE" ]; then
   echo "File not found ${FILE}"
```

Grueslayer:
Fixes #1212 by 
- setting MSWINHOME environment variable (to calculated $HOME from vim)
- exports these to WSL (if needed)
- using MSWINHOME in preview.sh for ~
- translating Windows path to WSL path

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -156,10 +156,8 @@ function! fzf#vim#with_preview(...)
     if empty($MSWINHOME)
       let $MSWINHOME = $HOME
     endif
-    if is_wsl_bash
-      if $WSLENV !~# '[:]\?MSWINHOME\(\/[^:]*\)\?\(:\|$\)'
-        let $WSLENV = 'MSWINHOME/u:'.$WSLENV
-      endif
+    if is_wsl_bash && $WSLENV !~# '[:]\?MSWINHOME\(\/[^:]*\)\?\(:\|$\)'
+      let $WSLENV = 'MSWINHOME/u:'.$WSLENV
     endif
     let preview_cmd = 'bash '.(is_wsl_bash
     \ ? substitute(substitute(s:bin.preview, '^\([A-Z]\):', '/mnt/\L\1', ''), '\', '/', 'g')
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -153,10 +153,10 @@ function! fzf#vim#with_preview(...)
   endif
   if s:is_win
     let is_wsl_bash = exepath('bash') =~? 'Windows[/\\]system32[/\\]bash.exe$'
+    if empty($MSWINHOME)
+      let $MSWINHOME = $HOME
+    endif
     if is_wsl_bash
-      if empty($MSWINHOME)
-        let $MSWINHOME = $HOME
-      endif
       if $WSLENV !~# '[:]\?MSWINHOME\(\/[^:]*\)\?\(:\|$\)'
         let $WSLENV = 'MSWINHOME/u:'.$WSLENV
       endif
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -153,6 +153,14 @@ function! fzf#vim#with_preview(...)
   endif
   if s:is_win
     let is_wsl_bash = exepath('bash') =~? 'Windows[/\\]system32[/\\]bash.exe$'
+    if is_wsl_bash
+      if empty($MSWINHOME)
+        let $MSWINHOME = $HOME
+      endif
+      if $WSLENV !~# '[:]\?MSWINHOME\(\/[^:]*\)\?\(:\|$\)'
+        let $WSLENV = 'MSWINHOME/u:'.$WSLENV
+      endif
+    endif
     let preview_cmd = 'bash '.(is_wsl_bash
     \ ? substitute(substitute(s:bin.preview, '^\([A-Z]\):', '/mnt/\L\1', ''), '\', '/', 'g')
     \ : escape(s:bin.preview, '\'))
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -12,7 +12,7 @@ IFS=':' read -r -a INPUT <<< "$1"
 FILE=${INPUT[0]}
 CENTER=${INPUT[1]}
 
-if [[ $1 =~ ^[A-Z]:\\ ]]; then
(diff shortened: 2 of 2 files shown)


bin/preview.sh
```diff
@@ -22,6 +22,20 @@ if [[ -n "$CENTER" && ! "$CENTER" =~ ^[0-9] ]]; then
 fi
 CENTER=${CENTER/[^0-9]*/}
 
+#MS Win support
+if [[ $FILE =~ "\\" ]]; then
```
junegunn, line 26: * Put a space after `#` for consistency
* Let's use single quotes to avoid escaping.

```sh
# MS Windows support
if [[ $FILE =~ '\' ]]; then
```

bin/preview.sh
```diff
@@ -22,6 +22,20 @@ if [[ -n "$CENTER" && ! "$CENTER" =~ ^[0-9] ]]; then
...
+  if [ -z "$MSWINHOME" ]; then
+    MSWINHOME="$HOMEDRIVE$HOMEPATH"
+  fi
+  if grep -qEi "(Microsoft|WSL)" /proc/version &> /dev/null ; then
+    MSWINHOME="${MSWINHOME//\\/\\\\}"
+    FILE="${FILE/#\~\\/$MSWINHOME\\}"
+    FILE=`wslpath -u "$FILE"`
+  elif [ ! -z "$MSWINHOME" ]; then
```
junegunn, line 34: `[ -n "$MSWINHOME" ]`

bin/preview.sh
```diff
@@ -22,6 +22,20 @@ if [[ -n "$CENTER" && ! "$CENTER" =~ ^[0-9] ]]; then
...
 
+#MS Win support
+if [[ $FILE =~ "\\" ]]; then
+  if [ -z "$MSWINHOME" ]; then
+    MSWINHOME="$HOMEDRIVE$HOMEPATH"
+  fi
+  if grep -qEi "(Microsoft|WSL)" /proc/version &> /dev/null ; then
+    MSWINHOME="${MSWINHOME//\\/\\\\}"
+    FILE="${FILE/#\~\\/$MSWINHOME\\}"
+    FILE=`wslpath -u "$FILE"`
```
junegunn, line 33: We prefer `$(...)` over backticks.

autoload/fzf/vim.vim
```diff
@@ -153,6 +153,14 @@ function! fzf#vim#with_preview(...)
   endif
   if s:is_win
     let is_wsl_bash = exepath('bash') =~? 'Windows[/\\]system32[/\\]bash.exe$'
+    if empty($MSWINHOME)
+      let $MSWINHOME = $HOME
+    endif
+    if is_wsl_bash
+      if $WSLENV !~# '[:]\?MSWINHOME\(\/[^:]*\)\?\(:\|$\)'
```
junegunn, line 160: Why don't we merge to conditions?

```vim
if is_wsl_bash && $WSLENV !~# '[:]\?MSWINHOME\(\/[^:]*\)\?\(:\|$\)'
```

junegunn (changes requested):

junegunn:
Thanks. I basically have no experience with Windows and I don't have a proper Windows/WSL environment to test this on, so I'll be asking for your help when an issue arises regarding the fix.

Issue #1050, opened 2020-06-11
Suggestion: include snippet descriptions in fuzzy search


katonori:
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

Why Snippets command searches words only in trigger words of snippets?
I think it's better to search words in snippet descriptions in addition to trigger words.

I modified the plugin like below by myself and it seems to work fine for me.

```
$ git diff 8f1e73b598d27d78dfb5843be19a73b6a3b222b1
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index 501101d..7706d49 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -914,7 +914,7 @@ function! fzf#vim#snippets(...)
   let colored = map(aligned, 's:yellow(v:val[0])."\t".v:val[1]')
   return s:fzf('snippets', {
   \ 'source':  colored,
-  \ 'options': '--ansi --tiebreak=index +m -n 1 -d "\t"',
+  \ 'options': '--ansi --tiebreak=index +m',
   \ 'sink':    s:function('s:inject_snippet')}, a:000)
 endfunction
```

Would you mind to change the default behavior to this or add an option to switch the behavior ?
I think many people likes this behavior better than current one.

junegunn:
I no longer use ultisnips so I don't have any opinion on it and I don't know if it's the right call to change the current behavior. I'll revisit the issue when there are more demands. But please note that just like any other commands in fzf.vim, you can easily override the default options like so:

```vim
command! -bar -bang Snippets
  \ call fzf#vim#snippets({'options': '--ansi --tiebreak=index +m'}, <bang>0)
```

MushiTheMoshi:
Hi @katonori, did you happen to workaround this problem? , I would like to know how to achieve this as well. BTW @junegunn  awesome work done with fzf, really productive tool. _

PR #1217, opened 2021-01-06
FIX: UltiSnips search only searches snippet value (not description) when using :Snippets command.

whatever555:
Before: (nothing is returned when searching by snippet description)   
image

  
Now (can search by both description and snippet):  
image
and  
image
  
  
Related issue: https://github.com/junegunn/fzf.vim/issues/1050


whatever555:
This has been open for a while with no feedback. Should I close this PR?

junegunn:
Sorry for the late reply. I don't use UltiSnips anymore so I haven't had a chance to test this. The patch looks good, but I would change it to something like `-n 1,..` so you can still perform anchored search e.g. `^foo`, `bar$`
What do you think?

whatever555:
apologies for the late reply:  yep, I can change that now.

autoload/fzf/vim.vim
```diff
@@ -964,7 +964,7 @@ function! fzf#vim#snippets(...)
   let colored = map(aligned, 's:yellow(v:val[0])."\t".v:val[1]')
   return s:fzf('snippets', {
   \ 'source':  colored,
-  \ 'options': '--ansi --tiebreak=index +m -n 1 -d "\t"',
+  \ 'options': '--ansi --tiebreak=index +m -n 1,2 -d "\t"',
```
whatever555, line 967: should that be ok ^

junegunn: No, I meant the literal `..` which means the whole range.

* `-n 1,..`: Let fzf search either on the first token only or on the whole string.

junegunn:
Merged, thanks.

PR #1223, opened 2021-01-15
Add a tag preview with a custom `tagpreview.sh` script

IngoMeyer441:


IngoMeyer441:
When the last line of the preview takes the full window width, the fzf scroll feature get activated (but there is no more line that could be scrolled to). Has anyone an idea why this happens?

junegunn:
I've tested the patch and it works as expected, thanks. I updated the branch to support `g:fzf_preview_window` configuration.

> fzf scroll feature gets activated

fzf does so in case the last line spans multiple lines when `--preview-window wrap` is set. That is not ideal, but I didn't want to make the implementation more complicated for the exceptional case.

junegunn:
Merged, thanks!

IngoMeyer441:
Hmm, strange, `wrap` is not set in my `FZF_DEFAULT_OPTS` but scrolling is activated any way.

bin/tagpreview.sh
```diff
@@ -0,0 +1,60 @@
...
+# Sometimes bat is installed as batcat.
+if command -v batcat > /dev/null; then
+  BATNAME="batcat"
+elif command -v bat > /dev/null; then
+  BATNAME="bat"
+fi
+
+if [ -z "$FZF_PREVIEW_COMMAND" ] && [ "${BATNAME:+x}" ]; then
+  ${BATNAME} --style="${BAT_STYLE:-numbers}" \
+             --color=always \
```
fabriziobertoglio1987, line 40: I was able to apply manually this patch to my fzf.vim 

1) I did not include the changes from https://github.com/junegunn/fzf.vim/pull/1223/commits/9067bcc4bfbba128c6c969700049c16d7dbe3485
2) I added to the above code to make it work with my

```
--theme="GitHub" \
```

junegunn: `bat` will use the theme set in `$BAT_THEME` environment variable so we don't need to change this code.

kkzhang:
not for windows

PR #1239, opened 2021-02-04
Add support to delete buffers in :Buffers command

mracos:
By pressing the CTRL-D keybinding.
Also enabled the `-m` multi select option to be able to delete multiple buffers.

---

Screen Shot 2021-02-03 at 23 46 44

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -688,18 +688,30 @@ function! s:bufopen(lines)
   if len(a:lines) < 2
     return
   endif
-  let b = matchstr(a:lines[1], '\[\zs[0-9]*\ze\]')
-  if empty(a:lines[0]) && get(g:, 'fzf_buffers_jump')
+
+  let key = a:lines[0]
+  let buffer_numbers = map(a:lines[1:], {_idx, bufline -> matchstr(bufline, '\[\zs[0-9]*\ze\]')})
+
+  if key == 'ctrl-d'
+    execute 'bdelete' join(buffer_numbers, ' ')
+    return
+  endif
+
+  let b = buffer_numbers[0]
+
+  if empty(key) && get(g:, 'fzf_buffers_jump')
     let [t, w] = s:find_open_window(b)
     if t
       call s:jump(t, w)
       return
     endif
   endif
-  let cmd = s:action_for(a:lines[0])
+
+  let cmd = s:action_for(key)
   if !empty(cmd)
     execute 'silent' cmd
   endif
+
   execute 'buffer' b
 endfunction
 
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -687,18 +687,30 @@ function! s:bufopen(lines)
...
+
+  let key = a:lines[0]
+  let buffer_numbers = map(a:lines[1:], {_idx, bufline -> matchstr(bufline, '\[\zs[0-9]*\ze\]')})
+
+  if key == 'ctrl-d'
+    execute 'bdelete' join(buffer_numbers, ' ')
+    return
+  endif
+
+  let b = buffer_numbers[0]
```
mracos, line 699: When multiple buffers are selected we open only the first one, not sure if this is the right/desired behaviour

dav-dev:
Would it be possible to also include Ctrl-W for buffer-wipe and also that possibility to delete/wipe the current (%) buffer which is currently unfortunately unselectable ?

mracos:
Hi @dav-dev!  I tried to make this patch as small as possible to increase the probability of being merged (easier to review, no big changes, etc). Let's first hear what the maintainers have to say and then we can improve upon (if gets merged haha). WDYT?

rameshsanth:
@junegunn Lets get this merged.. This is super useful for power users

@mracos   What do you think about "select all" buffers matching the search pattern for deletion ?

autoload/fzf/vim.vim
bruhtus: So we can delete multiple buffer at once? just like we open multiple files using tab in `:Files` or there's a different mechanism here?

Limero:
I was looking for this feature and found this PR. It's exactly what I'm missing, any potential to get it merged?

miklhh:
Sending my encouragement to get this exact functionality merged into fzf.vim. Would absolutely love it.

miklhh:
@mracos, I think it would be a really nice touch if `CTRL+D` did not close the `:Buffers` window, but instead closed the selected buffers and simply updated the `:Buffers` window. Further, when I tried your branch mracos:mf-add-delete-option-to-buffers-command, `:Buffers` no longer seem to honor my keybindings within the `:Buffers` window.

doronbehar:
> @mracos, I think it would be a really nice touch if `CTRL+D` did not close the `:Buffers` window, but instead closed the selected buffers and simply updated the `:Buffers` window. Further, when I tried your branch mracos:mf-add-delete-option-to-buffers-command, `:Buffers` no longer seem to honor my keybindings within the `:Buffers` window.

I too think this is the ideal behavior. I have been thinking about this feature for [a long time](https://github.com/junegunn/fzf/issues/2019#issuecomment-653084964), and after reading the changes here and the source code a bit, I think it'll require rather large rewrites to `function! s:fzf` and to `function! fzf#run` (from [junegunn/fzf](https://github.com/junegunn/fzf)).

The problem is that currently this PR uses the `--expect ctrl-d` fzf argument, and not `--bind=ctrl-d:execute(<cmd>)`. The hard part is to make `<cmd>` delete the line from the query, which is currently held in a temporary file when you run `:Buffers`.

The current implementation is great never the less, thanks for putting the hard work!

mracos:
Hey folks, thank you so much for the investigation and interest. It's been a while since I've last touched this PR and I have stopped using `fzf.vim` ever since 😞

So if anyone wants feel free to pick up this PR from there or even redo it from scratch, given that it's stale it might be easier

connermcd:
See PR above as a proposed solution. I have just a few comments on the thread:

> @mracos What do you think about "select all" buffers matching the search pattern for deletion ? I'm thinking add "Ctrl-a" to do select all and ctrl-d to delete them

While it is possible and relatively easy to use fzf's `--bind` to set `ctrl-a:select-all`. I did not include this in the PR as this behavior is inconsistent across fzf.vim and in other areas `alt-a:select-all` is used. I think it would be useful to have a consistent behavior for all functions that use `--multi`.

> Would it be possible to also include Ctrl-W for buffer-wipe and also the possibility to delete/wipe the current (%) buffer which is currently unfortunately unselectable ?

Adapting this PR to allow CTRL-W to execute `:bwipeout` is trivial to implement, but as the vim `:help :bwipeout` suggests you should only do this "if you know what you're doing". Usually `:bdelete` is sufficient and should be the more common use case, so I didn't implement this. Deleting or wiping the current buffer is as simple as typing `:bd` or `:bw` and I did not feel this needed to be an additional functionality within the plugin.

doronbehar:
BTW I use https://github.com/ibhagwan/fzf-lua which has this feature.

junegunn:
Thanks for closing this. The current version supports deletion via `CTRL-ALT-X`, allowing you to delete multiple buffers without closing the selector.

Issue #1261, opened 2021-03-13
Tag preview is hardcoded to use vim


gnfisher:
- [x] I have fzf 0.23.0 or above
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

The `tagpreview.sh` script is hardcoded to use `vim`. In my case, I only had neovim installed and was getting an error. If desired, we could check for `vim` and fallback to `nvim` as an alternative in the shell script.

PR #1262, opened 2021-03-13
Use nvim if vim not installed

gnfisher:
The `tagpreview.sh` script is hardcoded to the `vim` binary. For users that only
have  Neovim installed an error is displayed and no preview is rendered.

This change addresses this by falling back to the `nvim` binary if the `vim`
binary isn't present.

Addresses #1261

### DIFF
--- a/bin/tagpreview.sh
+++ b/bin/tagpreview.sh
@@ -20,7 +20,14 @@ if [ ! -r "$FILE" ]; then
   exit 1
 fi
 
-CENTER="$(vim -i NONE -u NONE -e -m -s "${FILE}" \
+# If users aren't using vim, they are probably using neovim
+if command -v vim > /dev/null; then
+  VIMNAME="vim"
+elif command -v nvim > /dev/null; then
+  VIMNAME="nvim"
+fi
+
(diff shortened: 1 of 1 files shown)


bin/tagpreview.sh
```diff
@@ -20,7 +20,14 @@ if [ ! -r "$FILE" ]; then
   exit 1
 fi
 
-CENTER="$(vim -i NONE -u NONE -e -m -s "${FILE}" \
+# If users aren't using vim, they are probably using neovim
+if command -v vim > /dev/null; then
+  VIMNAME="vim"
+elif command -v nvim > /dev/null; then
+  VIMNAME="nvim"
+fi
```
junegunn, line 28: Thanks.

Why don't exit here with an error message if both are unavailable?

junegunn:
Merged, thanks!

mckellygit:
Hi,
I have both vim and nvim installed.
Perhaps we should check $EDITOR for ~vim and if so use what that is set to ?
Or check the name of the parent process ?
thx,
-m

junegunn:
@mckellygit Hi, thanks for the suggestion. The command is only used to find the line number for the tag and it will give the same result on Vim and Neovim, so it doesn't matter which one we choose.

PR #1272, opened 2021-04-08
Add :DBuffers command

oliverralbertini:
This works like :Buffers, but allows you to select multiple buffers for
deletion. This is convenient when the listed buffers gets too high, and
you'd like to get rid of some.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -702,6 +702,16 @@ function! s:bufopen(lines)
   execute 'buffer' b
 endfunction
 
+function! s:bufdelete(lines)
+  if len(a:lines) < 2
+    return
+  endif
+  for line in a:lines[1:]
+    let b = matchstr(line, '\[\zs[0-9]*\ze\]')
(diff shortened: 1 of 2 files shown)


dav-dev:
isn't there already #1239 which looks very similar ?

oliverralbertini:
> isn't there already #1239 which looks very similar ?

Yes, I saw that after making this one!

arsham:
I prefer this version. It's easier to have two different key bindings for switching/deleting buffers.

TC-mSundaram:
Well this may never get merged, but the great thing about open source is that I can still use it locally. Thanks @oliverralbertini :sunglasses:

maheshsundaram:
Well this may never get merged, but the great thing about open source is that I can still use it locally. Thanks @oliverralbertini :sunglasses:

junegunn:
The current `:Buffers` command now supports deletion via `CTRL-ALT-X`, allowing you to delete multiple buffers without closing the selector.

PR #1274, opened 2021-04-17
[Snippets]Hide normal messages when inserting snippet

tarrex:
For better view, hide `=UltiSnips#ExpandSnippet()` message when inserting a snippet successfully. If the function calling does not succeed, the message will be displayed.

Before:
before


liskin:
Can you please try if https://github.com/junegunn/fzf.vim/pull/1196 works for you, as it also prevents these messages, and additionally also fixes several issues around snippets.

tarrex:
Yeah. It works fine for me, thanks for your code, I will close my PR. Hopefully your code will be merged into the master branch soon. 😄

PR #1289, opened 2021-05-25
[[B]Commits] Preserve window view

Frederick888:
In 161da95, [B]Commits were changed to range commands with -range=%, now
Vim jumps to the first line when these two commands are called directly
without specifying range.

This fixes it by saving window view in a temporary variable and
restoring it later. Reference: [1].

[1] https://vi.stackexchange.com/a/6037/33583

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1267,10 +1267,18 @@ function! s:given_range(line1, line2)
 endfunction
 
 function! fzf#vim#commits(...) range
+  if exists('b:fzf_winview')
+    call winrestview(b:fzf_winview)
+    unlet b:fzf_winview
+  endif
   return s:commits(s:given_range(a:firstline, a:lastline), 0, a:000)
 endfunction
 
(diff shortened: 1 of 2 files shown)


autoload/fzf/vim.vim
```diff
@@ -1267,10 +1267,14 @@ function! s:given_range(line1, line2)
 endfunction
 
 function! fzf#vim#commits(...) range
+  call winrestview(b:fzf_winview)
```
junegunn, line 1270: We should guard this code with `if exists('b:fzf_winview')`.

A user may have a custom version of the command directly calling the autoload function and we don't want it to fail after this change.

```sh
command! MyCommits call fzf#vim#commits({'left': '40%', 'options': '--preview-window up'})
```

junegunn (changes requested):

autoload/fzf/vim.vim
Frederick888: Pushed. Thanks for reviewing :)

junegunn:
Merged, thanks!

PR #1300, opened 2021-06-16
Add support for nushell

nathom:
Because of a quirk of the nushell syntax, it doesn't allow programs surrounded in quotes to be executed, as they are interpreted as string objects. I fixed this issue by making `bash` call the script.


danjenson:
same problem^ thanks for the solution!

junegunn:
Thanks!

PR #1326, opened 2021-09-10
Ensure bash scripts to always use unix line endings

wangzq:
Otherwise on Windows it may cause preview.sh to have CRLF as line endings and fail to preview.


junegunn:
Thanks!

PR #1337, opened 2021-10-10
Fix preview failing on files with special characters

notpeelz:
This fixes the preview script failing when given funky file names such as:
- `--whatever`
- `*`
- `file name`


junegunn:
Merged, thanks!

PR #1357, opened 2022-01-16
[Helptags] add preview window to :Helptags command

phlip9:
This PR adds a preview window to the vim help searcher.

+ In `fzf#vim#helptags`, we now also parse out the ex-command in the perl script so it's available for the preview (this is just an extra `\t(.*)` appended to the original regex).

+ Add the `--tag` placeholder to `:Helptags` with corresponding arguments.

And a beautiful screenshot of it in action : )

Screenshot 2022-01-15 204638


junegunn:
Merged, thanks!

PR #1362, opened 2022-01-29
[[B]Commits] Use --color=always with delta as well

liskin:
With some delta configurations (e.g. color-only = true, diff-highlight =
true), delta doesn't add its own coloring and keeps that from git diff,
and thus we need to turn that on.

Additionally, delta doesn't do its own detection of moved lines and
relies on git diff's colorMoved, which also needs enabling of colors.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1230,9 +1230,9 @@ function! s:commits(range, buffer_local, args)
   endif
 
   if !s:is_win && &columns > s:wide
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks. Can you post screenshots before and after the patch where this makes difference?

liskin:
Sure.

Without patch:
2022-01-30-222122_1915x1044_scrot

With patch:
2022-01-30-221956_1915x1044_scrot

My git/delta config is [here](https://github.com/liskin/dotfiles/blob/home/.config/git/config) if you need to dig deeper :-)

junegunn:
Merged, thanks.

PR #1366, opened 2022-02-03
Try to fix nvim windows wsl bash preview problem

NoahELE:
Sorry if this bothers you, because I actually know very little about viml. But this fix solves my problem.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -38,13 +38,10 @@ let s:bin = {
 \ 'tags':    s:bin_dir.'tags.pl' }
 let s:TYPE = {'dict': type({}), 'funcref': type(function('call')), 'string': type(''), 'list': type([])}
 if s:is_win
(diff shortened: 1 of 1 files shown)


junegunn:
Related: #557

@janlazo Can you drop the nvim branch as suggested here?

codito:
Confirming this fix works on nvim in Windows.

harrisonmg:
I have the same issue, and this fix works for me.

Running nvim in PowerShell with WSL bash installed.

```powershell
PS C:\Users\Harrison> cmd.exe /s /c 'for %A in ("C:\Users\Harrison\.local\share\nvim\plugged\fzf.vim/bin/preview.sh") do @echo %~sA'
C:\Users\Harrison\LOCAL~1\share\nvim\plugged\fzf.vim\bin\preview.sh
```

It looks like the 8.3 name doesn't like the leading `.` in `.local`.

Issue #1029, opened 2020-05-15
fzf-complete-word should respect dictionary option


nkcfan:
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

Currently fzf-complete-word is using hardcoded dictionary file `/usr/share/dict/words`.
https://github.com/junegunn/fzf.vim/blob/a74605b81d0dc2b28d35d9a1ab414dbdcbe3f45b/autoload/fzf/vim/complete.vim#L49-L53

It does not support vim option (a list of specified dictionary files)
```
set dictionary+=~/.vim/dict/words
```

reportaman:
+1

reportaman:
@nkcfan @junegunn 

I don't think any change is needed. This should work:

```vim
 function! fzf#vim#complete#word(...) 
   return fzf#vim#complete(s:extend({ 
     \ 'source': 'cat ' . &dictionary}, 
     \ get(a:000, 0, fzf#wrap()))) 
 endfunction 
```

PR #1373, opened 2022-03-02
[fzf#vim#complete#word]: respect dictionary option

jiz4oh:
related: https://github.com/junegunn/fzf.vim/issues/1029

fzf could respect dictionary option by a few changes. And fzf will maintain the behavior of the previous version if user don't have dictionary option by default.

### DIFF
--- a/autoload/fzf/vim/complete.vim
+++ b/autoload/fzf/vim/complete.vim
@@ -47,9 +47,9 @@ else
 endif
 
 function! fzf#vim#complete#word(...)
-  let sources = empty(&dictionary) ? '/usr/share/dict/words' : substitute(&dictionary, ',', ' ', 'g')
+  let sources = empty(&dictionary) ? ['/usr/share/dict/words'] : split(&dictionary, ',')
   return fzf#vim#complete(s:extend({
-    \ 'source': 'cat ' . sources},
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim/complete.vim
```diff
@@ -47,8 +47,9 @@ else
 endif
 
 function! fzf#vim#complete#word(...)
+  let sources = empty(&dictionary) ? '/usr/share/dict/words' : substitute(&dictionary, ',', ' ', 'g')
   return fzf#vim#complete(s:extend({
-    \ 'source': 'cat /usr/share/dict/words'},
+    \ 'source': 'cat ' . sources},
```
junegunn, line 52: Since we don't know what's inside the user's `&dictionary`, we have to `shellescape` each entry.

```vim
let sources = empty(&dictionary) ? ['/usr/share/dict/words'] : split(&dictionary, ',')

... 'cat ' . join(map(sources, 'shellescape(v:val)'))
```

jiz4oh: it sounds good to me, let's go with the changes.

junegunn (approved):

junegunn:
Merged, thanks!

PR #1375, opened 2022-03-09
Disable unicode for windows

iaalm:
Hi, I use fzf.vim on Windows with `ambiwidth=single`. When I run any long operations like "Rg", my vim keeps bell and there's no way to disable it from vim side. It's quite annoying. It also doesn't matter whether I use cmd or powershell. I find it related to this [issue](https://github.com/junegunn/fzf/issues/2631) because I don't see the spinner characters. Instead I see characters like "&" or "8" and the bell has same rhythm with the chars.
Screenshot

My suggestion is to add "--no-unicode" to all windows machine. Or maybe we can add a flag to disable spinner chars only?
Tested on Windows 11, gVim 8.2.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -158,7 +158,7 @@ function! fzf#vim#with_preview(...)
   if len(placeholder)
     let preview += ['--preview', preview_cmd.' '.placeholder]
   end
-  if &ambiwidth ==# 'double'
+  if &ambiwidth ==# 'double' || s:is_win
     let preview += ['--no-unicode']
   end
 


junegunn:
I'm not really a Windows user, so it's hard to tell if it's the right call to disable it altogether. Some systems are capable of properly rendering the spinner (see https://github.com/junegunn/fzf/issues/2631).

I'd suggest that you set it in your `$FZF_DEFAULT_OPTS` for the moment until we find a better way.


```vim
" In your vimrc
if has('win32') || has('win64')
  let $FZF_DEFAULT_OPTS = '--no-unicode'
endif
```

Issue #802, opened 2019-04-23
Show Current File BufferCommit Changes in Preview Window


neerajbadlani:
- Category
    - [ ] Question
    - [ ] Bug
    - [x] Suggestion
- OS
    - [x] Linux
    - [ ] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [x] Vim
    - [ ] Neovim

During BufferCommits , in the preview window , I see starting of the diff , is it possible to show just the diff of the current file in the commit . 
<!--

junegunn:
I was able to implement this like so:

```diff
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index ec1c00d..5df5bbc 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1128,14 +1128,16 @@ function! s:commits(buffer_local, args)
   \   '--expect=ctrl-y,'.expect_keys]
   \ }
 
+  let preview_command = 'git show --format=format: --color=always {SHA}'
   if a:buffer_local
     let options.options[-2] .= ', '.s:magenta('CTRL-D', 'Special').' to diff'
     let options.options[-1] .= ',ctrl-d'
+    let preview_command .= ' ' . fzf#shellescape(current)
   endif
 
   if !s:is_win && &columns > s:wide
     call extend(options.options,
-    \ ['--preview', 'echo {} | grep -o "[a-f0-9]\{7,\}" | head -1 | xargs git show --format=format: --color=always | head -200'])
+    \ ['--preview', 'echo {} | grep -o "[a-f0-9]\{7,\}" | head -1 | xargs -I {SHA} '.preview_command.' | head -200'])
   endif
 
   return s:fzf(a:buffer_local ? 'bcommits' : 'commits', options, a:args)
```

But on the second thought, I'm not quite sure if this behavior is always desirable. Sometimes, I want to see related changes in other files as well. It would be better if there's a way to just make the diff of the current file appears first.

neerajbadlani:
Actually this works pretty well for me , Possible to add -U8 flag to preview command in case u decide to commit this : git show -U8 --format=format: --color=always {SHA} ? .. Helps me to understand how the source code file evolved .

neerajbadlani:
Just 1 more little thing ... 

![image](https://user-images.githubusercontent.com/3440085/58703322-222bbe00-835d-11e9-80c1-d7aa92152ebf.png)

[...]

PR #1402, opened 2022-06-27
Show current file first in (B)Commits preview

Emilv2:
proposed solution for #802

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1230,8 +1230,10 @@ function! s:commits(range, buffer_local, args)
 
   if !s:is_win && &columns > s:wide
     let suffix = executable('delta') ? '| delta --width $FZF_PREVIEW_COLUMNS' : ''
+    let orderfile = tempname()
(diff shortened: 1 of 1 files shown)


autoload/fzf/vim.vim
```diff
@@ -1230,8 +1230,10 @@ function! s:commits(range, buffer_local, args)
 
   if !s:is_win && &columns > s:wide
     let suffix = executable('delta') ? '| delta --width $FZF_PREVIEW_COLUMNS' : ''
+    let s:orderfile = tempname()
+    call writefile([current], s:orderfile)
```
junegunn, line 1234: Is there any reason you use `s:` variable instead of a plain local variable?

Emilv2: No, I copied that from somewhere else in your script since I'm not very familiar with vimscript. I will remove it.

junegunn (approved):

junegunn:
Merged, thanks!

PR #1410, opened 2022-07-29
New g:fzf_quickfix_open option

ggustafsson:
Hello!

This patch add an option to disable the `copen` command inside of `fill_quickfix()` that several commands use. The default behaviour is good but it can get a bit weird when you already have several splits open (and I really like seeing just the `(1 out of X): info` status message) so I want to be able to disable it.

See it in action here: https://asciinema.org/a/511682

I just added a `if` statement with `get()` and tried to make everything fit in well. The code speaks for itself.

Best regards,
Göran Gustafsson


ggustafsson:
I'm closing this one because I can see this same behaviour Nvim + LSP stuff. Better to just make it easy to close location window for me...

PR #1427, opened 2022-09-21
Add clp support

jpe90:
This PR adds support for [clp](https://github.com/jpe90/clp) as an alternative to `bat` for file previewing.

In this implementation, `clp` is always used for file previewing if it's found. It assumes if the program is installed, the user would want to use it for file previewing, since that's its only function.

Alternatively a configuration option could be used to let users opt-in to `clp` more explicitly. Let me know if you would prefer doing it that way, or taking some other approach.

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -52,6 +52,10 @@ if [ -z "$CENTER" ]; then
   CENTER=0
 fi
 
+if command -v clp > /dev/null; then
+	CLPNAME="clp"
+fi
+
 # Sometimes bat is installed as batcat.
 if command -v batcat > /dev/null; then
   BATNAME="batcat"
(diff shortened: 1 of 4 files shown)


bin/preview.sh
```diff
@@ -52,14 +52,23 @@ if [ -z "$CENTER" ]; then
   CENTER=0
 fi
 
+if command -v clp > /dev/null; then
+	CLPNAME="clp"
+fi
+
 # Sometimes bat is installed as batcat.
 if command -v batcat > /dev/null; then
```
junegunn, line 60:
```suggestion
if command -v clp > /dev/null; then
  CLPNAME="clp"
elif command -v batcat > /dev/null; then
  # Sometimes bat is installed as batcat.
```

junegunn (commented):
Can clap show line numbers a la `bat --style=numbers`? I'd like to provide a consistent experience to the users regardless of the highlighter backend. We could use `clp` with `nl`, but then the numbers aren't highlighted.

jpe90:
@junegunn not currently, but that sounds like a good idea to me. I'll work on it.

jpe90:
@junegunn I noticed that bat line numbers don't seem to have background highlighting when they're on a highlighted line. 

Just to clarify, were you referring to the grey coloring of the number line text?

Screen Shot 2022-09-22 at 2 21 43 PM

junegunn:
> Just to clarify, were you referring to the grey coloring of the number line text?

Yes. With `clp SOME_FILE | nl`, we don't get highlighted line numbers.

Issue #1432, opened 2022-10-11
fzf.vim cannot be used on Windows inside git-bash shell. [Interested to push a PR, but I need many directions]


Adnn:
- [x] I have fzf 0.30.0 or above
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [ ] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

I have been using fzf successfully under Ubuntu bash shell. 

But there are several issues when trying to use in with git-bash shell on Windows 10.

## Without `set shell`
If I do not set the shell in my `init.vim`, the shell is:

    shell="C:\Program Files\Git\usr\bin\bash.exe"

Then I cannot get any `fzf` command to execute. E.g.

     :Files

fails with error:

> Error detected while processing function fzf#vim#files[15]..<SNR>94_fzf[1]..<SNR>94_check_requirements[11]..fzf#exec[31]..function fzf#vim#files[15]..<SNR>94_fzf[1]..<SNR>94_check_requirements[11]..fzf#exec:
> line   31:
> E605: Exception not caught: Failed to run "C:\Users\adnn\AppData\Local\nvim-data\plugged\fzf/bin/fzf --version"

## With `set shell=cmd`

`fzf` commands `:Files` and `:Rg` work, as they show the pop-up window that is searchable. But the preview does not work:

 > bash: C:UsersadnAppDataLocalNVIM-D~1pluggedfzf.vimbinpreview.sh: No such file or directory

`:Buffers` does not work at all. There is an error showing in the pop-up window

> [Command failed: type ^"C:\Users\adn\AppData\Local\Temp\nvim.0\FCdV6q\4^"]

---

I would be interested to address these issues, but I never worked with vim script, nor this project sources, so I would need directions.
I am most interested to address the first case (without setting a shell, which then uses git's `bash.exe`). From the error, I suppose the problem is that `bash` cann execute `fzf - -version` because the path starting with `C:\Users` is not a valid git-bash path.

Any pointer where this path is generated in the plugin code?

Adnn:
I created the PR above #1434.
This addresses the case of `shell=cmd` issue with the preview pane.

I am happy to get feedback and rework the PR to get it accepted.

3N4N:
[...]

PR #1434, opened 2022-10-12
Fix preview under Neovim for Windows (running in git-bash).

Adnn:
Please see #1432


junegunn:
Thanks for looking into the issue. Unfortunately, I'm not a Windows user and I'm not confident in reviewing this. Most of the work on Windows support was previously done by other contributors. Maybe @Melandel or @janlazo can give some help here?

PR #1441, opened 2022-12-02
bin/preview.sh: Print content based on context position

dodgerblue:
bat / batcat supports a `line-range` parameter to specify the range of the file to display. Set the start position based on the provided CENTER variable and FZF_DEFAULT_LINES. The preview window will contain only the lines around the CENTER, not the beginning of the file.

For the default awk-based approach, only print the lines that are greater than the start point.

Signed-off-by: Bogdan Purcareata <bogdan.purcareata@gmail.com>

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -27,6 +27,8 @@ if [[ -n "$CENTER" && ! "$CENTER" =~ ^[0-9] ]]; then
   exit 1
 fi
 CENTER=${CENTER/[^0-9]*/}
+START=$(($CENTER - $(($FZF_PREVIEW_LINES / 2))))
+[[ $START -lt 0 ]] && START=0
 
 # MS Win support
 if [[ "$FILE" =~ '\' ]]; then
@@ -61,7 +63,7 @@ fi
 
 if [ -z "$FZF_PREVIEW_COMMAND" ] && [ "${BATNAME:+x}" ]; then
   ${BATNAME} --style="${BAT_STYLE:-numbers}" --color=always --pager=never \
-      --highlight-line=$CENTER -- "$FILE"
+      --highlight-line=$CENTER --line-range=$START: -- "$FILE"
   exit $?
 fi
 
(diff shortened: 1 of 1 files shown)


junegunn:
We used to use `--line-range` in the past, but it was removed in 9ceac718026fd39498d95ff04fa04d3e40c465d7, which made it possible to scroll upwards. To allow scrolling, the script now prints the whole file.

dodgerblue:
Thank you for the explanation and the reference, @junegunn ! Would you be willing to consider a change where `--line-range` is used based on an environment variable? Personally, I'm more interested in getting the quick glimpse of the location in the preview, rather than scrolling towards it. I'll also look into 'pre-scrolling' the window programatically to the center when displaying the preview.

junegunn:
I'm not sure what you're trying to do here. The preview window of fzf already scrolls to the line and display the line at the center of the screen (e.g. `'--preview-window', '+{2}-/2'`). Ag, Rg, Buffers, BTags, etc.

dodgerblue:
Ah, thank you for the pointer! I looked into the `Ag` & co definitions and got what you mean. It's what I had in mind when I mentioned 'pre-scroll'.

In my use-case, I'm defining a new command via `fzf#run` for browsing tags, but this one's based on [GNU global](www.gnu.org/software/global/). For this, I'm using `source: global -cT`, which simply prints the symbol names matching a prefix. When displaying the preview, I'm doing some extra processing to figure out the file and line, and I'm using `with_preview('placeholder': "$(global -t {} | tail -n 1 | ...)")`. This works fine when displaying the preview window contents, but unfortunately I don't have the file and line ready before running the fzf command, so that I can pass the scroll position to `--preview-window`. Changing the `source` command to include this extra processing adds some delay to the command (tested only in terminal, haven't found a way to make it work with `fzf#run` yet, with all the pipes and escapes), and I can't seem to be able to replace `+{2}-/2` with an expression like `+{$( ... )}-/2` or `+$( ... )-/2` to dynamically compute the position.

I'll look more into it. Maybe there's a fzf feature that supports some extra processing of the entry before it's sent to the preview window, that I don't know about?

junegunn:
> Changing the `source` command to include this extra processing adds some delay to the command (tested only in terminal, haven't found a way to make it work with `fzf#run` yet, with all the pipes and escapes), and I can't seem to be able to replace `+{2}-/2` with an expression like `+{$( ... )}-/2` or `+$( ... )-/2` to dynamically compute the position.

I see, thanks for the clarification.

> Maybe there's a fzf feature that supports some extra processing of the entry before it's sent to the preview window

No, there's no such thing.

PR #1447, opened 2022-12-31
fix: GFiles is ignoring args

ksmai:
The commit linked below removed `a:args` from the `git ls-files` command, making it unable to take any options like `--cached`, `--others` and `--exclude-standard`

https://github.com/junegunn/fzf.vim/commit/531dd67350331771c0c099e9682a05744149e7b9


junegunn:
Merged, thanks!

PR #1455, opened 2023-01-25
[[B]Commits] Support arguments for git-log

druckdev:
Add support for command arguments that are passed to git-log (e.g. `:Commits -S foo -3`).

The place of insertion is relevant. Generally the arguments should come at the end to overwrite existing flags if necessary, but need to come before the pathspec (i.e. `fzf#shellescape(current)`) in BCommits.


autoload/fzf/vim.vim
```diff
@@ -1320,6 +1322,7 @@ function! s:given_range(line1, line2)
   return []
 endfunction
 
+" [[args], fullscreen-boolean, [git-log-args]]
```
junegunn, line 1325: fullscreen boolean should be the last argument by convention.

druckdev: Oh good to know.

Then I would move `<q-args>` to the front so that it can be easily separated before the rest is passed to `s:fzf`. What do you think?

junegunn:
Now all three parameters are optional, we have to make sure all possible combinations work as expected. See the commit message for the list of commands and function calls I have tested with the updated code.

Please let me know if something is not working as expected or if there's anything else you want to address.

druckdev:
LGTM, thanks!

Also for changing to `join()`. I am not so versed in vim-script and did not think/know of that.

druckdev:
What do you think of adding `-complete=file` to Commits? For BCommits it does not really make sense, but I see a use-case in specifying files or folders to  Commits.

One would have to reformat the `s:defs` block then...

junegunn (approved):

junegunn:
Merged, thanks!

PR #1462, opened 2023-03-13
docs: fix regex searches with shellescape for the query

alexandervantrijffel:
Fix regex regex searches in the example for Advanced ripgrep integration. Without this, regex searches are not functioning.


junegunn:
> regex searches

Such as?

When `options` entry is given as an array like this, fzf "shell-escape"s each of them automatically, so no manual shellescape is required.
https://github.com/junegunn/fzf/blob/1116e481bef4f63abe0e506ce20469e244c505e6/plugin/fzf.vim#L374-L377

PR #1465, opened 2023-03-29
#1239 Add ability to bdelete buffers with CTRL-D from :Buffers command

connermcd:
This pull request extends the :Buffers command with `--multi` to allow `<Tab>`
selection of multiple buffers that can then be deleted with CTRL-D or
opened into separate tabs/splits/vertical splits with
CTRL-T/CTRL-X/CTRL-V. See #1239 and #1272.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -739,7 +739,15 @@ function! s:bufopen(lines)
   if len(a:lines) < 2
     return
   endif
+
   let b = matchstr(a:lines[1], '\[\zs[0-9]*\ze\]')
+
+  if a:lines[0] == 'ctrl-d'
+    execute 'silent bdelete' b
+    call fzf#vim#buffers()
+    return
+  endif
+
   if empty(a:lines[0]) && get(g:, 'fzf_buffers_jump')
     let [t, w] = s:find_open_window(b)
     if t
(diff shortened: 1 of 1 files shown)


zatchheems:
just bumping this to show interest for this feature! it's something I've been wanting for a year or so and would use on a daily basis. (yes, even nearly 5 years after #1239 was opened 😁)

if this has already been implemented and I'm out of the loop please correct me.

DanSM-5:
@zatchheems It has not been implemented yet. I've been waiting for that too.

I have a hack around `fzf#vim#buffers` to make this possible but it is fragile and I'd prefer to see this merged. However, if you are interested (while we wait), see the [function](https://github.com/DanSM-5/vim-config/blob/38f187b1c6eedef2019c730db6a1a4fb1c72babf/autoload/config.vim#L1238) and it's helper [script](https://github.com/DanSM-5/vim-config/blob/38f187b1c6eedef2019c730db6a1a4fb1c72babf/utils/remove_buff.sh).

zatchheems:
@DanSM-5 thanks for that. it's _verbose!_ would also love to see the feature merged here but I'll give your function a shot. it'll save me some keystrokes!

junegunn:
Apologies for the delayed response. I was looking for a way to delete buffers without closing the selector.

The current version now supports deletion via `CTRL-ALT-X`, allowing you to delete multiple buffers without closing the selector.

PR #1467, opened 2023-04-02
Update preview.sh to use bat's ability to jump to line range

dberlin:
When using bat, preview.sh will often end up trying to highlight lines that are off screen.

These days (since version 0.8, and it's on version 0.23, so a while), bat has the ability to specify a line range in the file to display.

We use it, plus the terminal height, to ensure the highlighted center line always visible.

We could use stty instead of tput to get terminal size, but everything i tried this on had tput available.

(Unfortunately, $LINES is only available in bash interactive shells)

Bat doesn't care if line-range specifies a range larger than the file, so TOPRANGE > number of lines in file is fine.

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -27,6 +27,9 @@ if [[ -n "$CENTER" && ! "$CENTER" =~ ^[0-9] ]]; then
   exit 1
 fi
 CENTER=${CENTER/[^0-9]*/}
+(("BOTTOMRANGE=$CENTER - ($(tput lines)/2)"))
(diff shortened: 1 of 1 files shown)


junegunn:
We used to use `line-range`, but it was removed in https://github.com/junegunn/fzf.vim/commit/4145f53f3d343c389ff974b1f1a68eeb39fba18b in favor or the new preview window offset expression which allows upward scroll.

```sh
cat /usr/share/dict/words | nl | fzf --preview 'cat /usr/share/dict/words' --preview-window '+{1}'
```

PR #1478, opened 2023-05-20
feat/add :changes command to the list

Sanix-Darker:
## WHAT

Added `:Changes` to the plugin to preview changes made inside a buffer and jump
on that line.

## WHY

Related to : https://github.com/junegunn/fzf.vim/issues/1479

## HOW

- feat: add :changes command from Vim as a fzf command to jump on the line of a change made
- doc: adapt the documentation

## DEMO

https://github.com/junegunn/fzf.vim/assets/22576758/ec44d9f9-c6eb-49e8-af07-c60bd9af9e4f

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1233,11 +1233,13 @@ function! fzf#vim#changes(...)
   redir => cout
   silent changes
   redir END
-  let list = split(cout, "\n")
+  let s:changelist = split(cout, "\n")
+  let current = -match(s:changelist, '\v^\s*\>')
   return s:fzf('changes', {
-  \ 'source':  extend(list[0:0], map(list[1:], 's:format_change(v:val)')),
+  \ 'source':  extend(s:changelist[0:0], map(s:changelist[1:], 's:format_change(v:val)')),
   \ 'sink*':   s:function('s:change_sink'),
-  \ 'options': '+m -x --ansi --tiebreak=index --header-lines 1 --tiebreak=begin --prompt "Changes> "'}, a:000)
+  \ 'options' : '+m -x --ansi --tiebreak=index --cycle --scroll-off 999 --sync --bind start:pos:'.current.' --tac --header-lines 1 --tiebreak=begin --prompt "Changes> "',
+  \ }, a:000)
 endfunction
 
 " ------------------------------------------------------------------
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1214,7 +1214,7 @@ endfunction
 " Changes
 " ------------------------------------------------------------------
 function! s:format_change(line)
-  return substitute(a:line, '\d\{1,2}', '\=s:yellow(submatch(0), "Number")', '')
+    return substitute(a:line, '\d\+', '\=s:yellow(submatch(0), "Number")', '')
 endfunction
 
 function! s:change_sink(lines)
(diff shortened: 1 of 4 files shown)


junegunn:
This might be of your interest: 5d87ac1fe8d729f116bda2f90a7211ad309ddf5a

Sanix-Darker:
Thanks for the review @junegunn , i updated the code with your recommendations here : https://https///github.com/junegunn/fzf.vim/pull/1478/commits/f3c4d87

autoload/fzf/vim.vim
```diff
@@ -1210,6 +1210,38 @@ function! fzf#vim#jumps(...)
   \ }, a:000)
 endfunction
 
+" ------------------------------------------------------------------
+" Changes
+" ------------------------------------------------------------------
+function! s:format_change(line)
+  return substitute(a:line, '\d\{1,2}', '\=s:yellow(submatch(0), "Number")', '')
```
junegunn, line 1217:
```suggestion
  return substitute(a:line, '\d\+', '\=s:yellow(submatch(0), "Number")', '')
```

autoload/fzf/vim.vim
```diff
@@ -1210,6 +1210,38 @@ function! fzf#vim#jumps(...)
...
+endfunction
+
+function! fzf#vim#changes(...)
+  redir => cout
+  silent changes
+  redir END
+  let s:changelist = split(cout, "\n")
+  let current = -match(s:changelist, '\v^\s*\>')
+  return s:fzf('changes', {
+  \ 'source':  extend(s:changelist[0:0], map(s:changelist[1:], 's:format_change(v:val)')),
```
junegunn, line 1239: Don't have to be an `s:` variable.

```suggestion
  let list = split(cout, "\n")
  let current = -match(list, '\v^\s*\>')
  return s:fzf('changes', {
  \ 'source':  extend(list[0:0], map(list[1:], 's:format_change(v:val)')),
```

Sanix-Darker: https///github.com/junegunn/fzf.vim/pull/1478/commits/b90f4d3

autoload/fzf/vim.vim
```diff
@@ -1210,6 +1210,38 @@ function! fzf#vim#jumps(...)
...
+function! s:change_sink(lines)
+  if len(a:lines) < 2
+    return
+  endif
+  let cmd = s:action_for(a:lines[0])
+  if !empty(cmd)
+    execute 'silent' cmd
+  endif
+  let linecol = split(a:lines[1])
+  execute 'normal! '.linecol[1].'Gzz'
```
junegunn, line 1229: * Doesn't work if the line starts with `>`: `>   0     1    8` ->  `0Gzz`
* We should also move the cursor to the designated column position. i.e. `call cursor(line, col)`
    * Note that we should not use `<COL>|` because it moves the cursor to the "screen column", not the real column number shown in `:changes`.

autoload/fzf/vim.vim
```diff
@@ -1210,6 +1210,38 @@ function! fzf#vim#jumps(...)
   \ }, a:000)
 endfunction
 
+" ------------------------------------------------------------------
+" Changes
+" ------------------------------------------------------------------
+function! s:format_change(line)
+    return substitute(a:line, '\d\+', '\=s:yellow(submatch(0), "Number")', '')
```
junegunn, line 1217:
```suggestion
  return substitute(a:line, '\d\+', '\=s:yellow(submatch(0), "Number")', '')
```

PR #1482, opened 2023-05-31
Remove dependencies of :Helptags on grep and sort to make it easier to get it working on Windows

edwinst:
This removes fzf#vim#helptags dependencies on external "grep" and "sort" commands. I think it also makes the code simpler. The special filename handling for Windows could be removed.

I put the removal of the external "sort" in a separate commit since it can affect the sort order. E.g. my Windows "sort" command puts single-quote (') before bang (!), while Perl's sort does the reverse. (Perl sorts these in ASCII order.)

Note that this does not make :Helptags fully work for me on Windows. There is one further problem: preview.sh is called with colons as separators for the filenames, etc., which breaks on Windows. I fixed it locally but I'm not sure what a good universal separating character other than colon could be. I used '$' locally but that may not be a good choice for Linux users.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1146,8 +1146,8 @@ function! s:helptag_sink(line)
 endfunction
 
 function! fzf#vim#helptags(...)
-  if !executable('grep') || !executable('perl')
-    return s:warn('Helptags command requires grep and perl')
+  if !executable('perl')
+    return s:warn('Helptags command requires perl')
   endif
   let sorted = sort(split(globpath(&runtimepath, 'doc/tags', 1), '\n'))
   let tags = exists('*uniq') ? uniq(sorted) : fzf#vim#_uniq(sorted)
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks.

But I'm getting different results with and without the patch

### without patch

image

### with patch

image

edwinst:
That's very strange. You seem to get almost 10x more results with the patch and still the lines starting with bang '!' are missing, although they should be the first ones.

I will try to repro that in Linux. Sorry about that.

edwinst:
On Linux I get different sort order but the same number of results. I created a new pull request that only replaces "grep" and keeps the external "sort", in order to keep the sort order. (Windows sort is pretty useless but Linux sort seems to give dictionary order by default which is kind of useful.)

https://github.com/junegunn/fzf.vim/pull/1483

This should at least help to narrow down the problem. I still don't know why you get a different number of results with the patch, unfortunately. Maybe you have a minute to test the new pull request. Thanks!

junegunn:
There was a bug in your Perl script. No duplicates now.

junegunn (approved):
LGTM now. Is there anything else you want to address?

edwinst:
D'oh, that was a stupid bug! I probably missed it because on my system the only large file came last, so I did not notice the duplications.

There's still the user-facing question of what is the preferred sort order. GNU sort seems to do some kind of dictionary order by default which might be perceived as more useful than the codepoint order produced by Perl's sort:

    $ (echo '!test' ; echo bar) | sort
    bar
    !test
    $ (echo '!test' ; echo bar) | perl -ne 'push @a,$_; END { print for sort @a }'
    !test
    bar

That's why I would have leaned towards #1483 , in order not to change possibly desired behavior for Linux users.

For Windows, the only other thing that keeps :Helptags from working is that the tagpreview.sh script expects some arguments, including the filename, to be separated by colons, which breaks on Windows. I'll try to find a robust fix for that, if my time permits.

junegunn:
The results are identical on my system, so I think we're good to go.
 
```sh
$ # macOS sort
$ (echo '!test' ; echo bar) | sort
!test
bar

$ # GNU sort
$ (echo '!test' ; echo bar) | gsort
!test
bar

$ (echo '!test' ; echo bar) | perl -ne 'push @a,$_; END { print for sort @a }'
!test
bar

```

junegunn:
Merged, thanks!

PR #1485, opened 2023-06-10
README - Improve installation guide

kevin-jones:
In addition to Plug, there are other popular plugin managers too and I spent quite a while trying to get fzf working as I didn't realise 2 plugins are required in vim. Online, I found quite a few posts on forums and blogs from people having similar trouble. Hopefully this documentation update helps others!

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ selector with fzf.
 Installation
 ------------
 
+To use this vim plugin, [another vim plugin][fzf-main] must also be installed first.
+
 fzf.vim depends on the basic Vim plugin of [the main fzf
 repository][fzf-main], which means you need to **set up both "fzf" and
 "fzf.vim" on Vim**. To learn more about fzf/Vim integration, see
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks but,

> image

(https://github.com/junegunn/fzf.vim/commit/c0a5fee7071ed89602b9b59fb67e83d6cd23addc)

Can we be more clear than this? This is the very first sentence in the installation section of the page, and we haven't talked about any plugin manager yet.

So I'm really doubtful that adding more instructions and making the page longer will help people who don't even read a single sentence in the documentation.

junegunn:
Also, this plugin will print this error message if the main Vim plugin is not installed.

https://github.com/junegunn/fzf.vim/blob/1dcdb21db618055134cd611f4f5918f6d00a5df0/autoload/fzf/vim.vim#L109-L111

image

kevin-jones:
> Can we be more clear than this?

You can, yes, that's why I've provided an update which follows the great examples set by all of the best README files for other vim plugins.

As the creator of Plug, perhaps you're reluctant, I don't know. It's unusual to have a vim plugin that relies on another vim plugin with almost the exact same name, especially when the other vim plugin isn't really just a vim plugin.

So - yes - you can be more clear. I try/install vim plugins all of the time - yours is the only one I got tripped up on and if you search online forums etc. there are a lots of other people that fell into the same trap as me.

Take it or leave it, up to you 🙂

junegunn:
>  I try/install vim plugins all of the time 

I need some clarification.

* Did you read the instructions before installing this plugin? My guess is that you have prior experience with other plugins so you just skipped it over thinking it would be the same?
* Did you read the instructions right after you failed to get it working? Or you read it after spending some time searching for the web for the resolution?
* Did you see the error message saying that the dependency is not met? i.e. `fzf#run function not found. You also need Vim plugin from the main fzf repository (i.e. junegunn/fzf *and* junegunn/fzf.vim)`


> there are a lots of other people that fell into the same trap as me.

The documentation and the error message weren't clear enough from the start. Maybe many of the posts you saw were written before the progress was made.

* https://github.com/junegunn/fzf.vim/commit/01f7146cf6e903e31cd900944393bec86c6da4a6
* https://github.com/junegunn/fzf.vim/commit/c0a5fee7071ed89602b9b59fb67e83d6cd23addc

I don't know which plugin manager you use, but I'm pretty sure you know how to install two plugins with it. And it's what the current documentation is telling you to do. I don't see much point in listing a growing set of commands for different plugin managers that are more or less the same.

PR #1487, opened 2023-06-29
Add preview for ":Windows"

felixendres:
I've added/fixed the preview for the :Windows command. Since I have preview enabled from an environment variable, all I saw in the preview pane was an error message. I'd say it is generally useful to have the preview for :Windows, so I'll offer this PR.


junegunn:
> Since I have preview enabled from an environment variable

I don't recommend the practice. Because fzf is a text filter program that is used for all kinds of things, and a single preview command can never handle all the different types of input.

https://github.com/junegunn/fzf#preview-window

---

Having said that, I've tested your patch but it doesn't properly handle an empty buffer named `[No Name]`.

PR #1488, opened 2023-07-15
tagpreview: open vim in readonly mode

nicovince:
When computing the center line, if the file is open the vim command returns with a non-zero exit code and the `return` is executed. This yields a weird error message "return: can only `return'" in the preview window, even if the preview is properly displayed

### DIFF
--- a/bin/tagpreview.sh
+++ b/bin/tagpreview.sh
@@ -30,10 +30,10 @@ else
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks. How about also changing `return` to `exit`? Because it's clearly wrong in the context.

nicovince:
> Thanks. How about also changing `return` to `exit`? Because it's clearly wrong in the context.

Yes, I updated the pull request.

junegunn:
Thanks.

PR #1495, opened 2023-08-10
[[B]Commits] Bump Git commit hash length limit to 40

Frederick888:
When working in large repositories, 9 characters can be ambiguous.
Bumping it to 40, which is the length of SHA-1. (The algorithm is not
that relevant though. Even when SHA-256 takes over, 40 characters should
still be enough for most repositories.)


junegunn:
LGTM, thanks!

Issue #1345, opened 2021-11-08
RG command should respect fzf_actions 


ViRu-ThE-ViRuS:
- [x] I have fzf 0.23.0 or above
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

RG commands should also respect fzf_actions dict. including function calls, splits, etc.

bruhtus:
Not sure about the function calls but the splits keybindings works for me with `:Rg` command too. I also remap the `g:fzf_action` to something like this (only changed the default vertical splits keybindings):
```vim
 let g:fzf_action = {
        \ 'ctrl-t': 'tab split',
        \ 'ctrl-x': 'split',
        \ 'ctrl-s': 'vsplit' }
```

Can you give more context how to reproduce the issue?

ViRu-ThE-ViRuS:
@bruhtus thanks for your prompt response, let me try to explain what was happening:

``` lua
local utils = require("utils")

vim.g.fzf_preview_window = "right:50%:+{2}-/2"
vim.g.fzf_buffers_jump = 1
vim.g.fzf_layout = {["down"] = "40%"}

vim.g.fzf_action = {
    ["ctrl-q"] = utils.qf_populate,
    ["ctrl-x"] = "split",
    ["ctrl-v"] = "vsplit"
}

utils.map("n", "<c-p>p", "<cmd>Files<cr>")
utils.map("n", "<c-p>b", "<cmd>Buffers<cr>")
utils.map("n", "<c-p>f", "<cmd>Rg<cr>")
utils.map("n", "<c-p>z", "<cmd>Rg TODO<cr>")

vim.cmd [[
    let $FZF_DEFAULT_COMMAND = 'rg --files --follow --smart-case --hidden --no-ignore -g "!{.DS_Store,.cache,venv,.git,.clangd,.ccls-cache}" 2> /dev/null'
    let $FZF_DEFAULT_OPTS = '--bind ctrl-a:select-all,ctrl-d:deselect-all'

    augroup FZF_UI
        autocmd!
        autocmd User FzfStatusLine setlocal statusline=%!v:lua.StatusLine('FZF')
    augroup end
]]
```

[...]

PR #846, opened 2019-07-17
Add configuration to switch between quickfix and loclist

mtzeng:

PR #1500, opened 2023-09-03
Add option to use location list instead of quickfix list

### LINKED ISSUE #432: ability to choose qf/location list
- Category
    - [ ] Question
    - [ ] Bug
    - [x] Suggestion
- Vim
    - [ ] Vim
    - [x] Neovim

I've started to write some functions to be able to choose a quickfix/location list among the ones loaded (visibler through :chistory) but it's not easy dealing with :chistory/:colder/:cnewer . Is there any existing implementation someone would know of ? If no, is that sthg that could make it to fzf.vim ?

### LINKED ISSUE #448: option for using location instead of quickfix for :Ag
- Category
    - [ ] Question
    - [ ] Bug
    - [X] Suggestion
- OS
    - [X] Linux
    - [ ] macOS
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [X] Vim
    - [X] Neovim

Hi, is it possible to add a option to use location instead of quickfix for the example shown in this issue: #185 ?

I'm currently using the following function to generate a location list:
```vimscript
function! s:build_location_list(lines)
  call setloclist(0, map(copy(a:lines), '{ "filename": v:val }'))
  lopen
  ll
endfunction
```
This does not work with `:Ag` hoewer, where i had to modifiy `autoload/fzf/vim.vim::631`

nkouevda:
Fixes #432, fixes #448.

Feel free to close this if you don't want it; I see the readme says this repo is "_not_ designed to be flexible or configurable", so totally understandable.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -374,13 +374,19 @@ function! s:warn(message)
   return 0
 endfunction
 
-function! s:fill_quickfix(list, ...)
+function! s:fill_list(list, ...)
   if len(a:list) > 1
-    call setqflist(a:list)
-    copen
+    let use_location_list = get(g:, 'fzf_use_location_list', 0)
+    if use_location_list
+      call setloclist(0, a:list)
+      lopen
+    else
+      call setqflist(a:list)
+      copen
+    endif
     wincmd p
     if a:0
-      execute a:1
+      execute (use_location_list ? 'l' : 'c') . a:1
     endif
   endif
 endfunction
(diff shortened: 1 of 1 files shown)


junegunn:
You might want to check out #846.

Can you explain why you want to (have to) use the location list instead of the quickfix list?

nkouevda:
Partially habit / convenience / consistency; I'm used to using location lists instead of the quickfix list, and navigating items with vim-unimpaired's `[l` / `]l` instead of `[q` / `]q`. e.g. I use [your `s:build_quickfix_list` example](https://github.com/junegunn/fzf/blob/0f50dc848e58b47408713c2c31f2be73287c030a/README-VIM.md#examples), but [using location list](https://github.com/nkouevda/dotfiles/blob/73e38daaf6a647f7419583fd9c304cdac7cdaaf5/.vim/vimrc#L287) instead. (But [as you explained](https://github.com/junegunn/fzf.vim/issues/1345#issuecomment-963743146) that doesn't work for `:Rg`, which is how I ended up here.)

And in terms of functionality, this allows having multiple lists, e.g. I have wanted to search for different things in different windows, and to iterate through the results separately.

But admittedly (1) is entirely subjective, and (2) is not a particularly common use case in my experience.

junegunn:
Thanks for the clarification.

> this allows having multiple lists

If that's the case, we should probably provide a way to use a different list type per command. I'll see what I can do.

junegunn:
See 832a090870a79d8be87d8856c821806512a45b23.

It's a pretty big change, so please give it a try and let me know how it works.

```vim
let g:fzf_vim = {}
let g:fzf_vim.listproc = { list -> fzf#vim#listproc#location(list) }
```

nkouevda:
Works great; thank you!

I don't use `preview_bash`, but noticed that needs a little fix: #1502

PR #1502, opened 2023-09-06
Use s:conf for preview_bash and files_options

nkouevda:
Follow-up to 832a090870a79d8be87d8856c821806512a45b23.


junegunn:
Nice find, thanks!

PR #1503, opened 2023-09-06
Use fzf#shellescape instead

jiz4oh:
Hi @junegunn, your code is exceptionally well-written. It's evident that you've put a lot of thought and effort into it. I'm particularly impressed by the clean and efficient structure you've created. The readability and organization make it a pleasure to work with. however, I have a small suggestion to enhance it even further. considering the way to work with external program, `fzf#shellescape` is better


junegunn:
Thanks! I found a couple more places where bare shellescape is used and changed them as well.

PR #1507, opened 2023-09-17
Replace unicode figure space (U+2007) with regular space

balki:
I assume this is not intentional as the doc file is a regular txt file. I see this in fzf repo as well.


junegunn:
Thanks. I'll update the file on the main repository as well.

Issue #1389, opened 2022-04-30
Feature Request: Add command for searching available man pages


AlxHnr:
- [ ] I have fzf 0.30.0 or above
- [X] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [X] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [X] I have read through the manual page of fzf (`man fzf`)
- [X] I have searched through the existing issues

It would be great to provide a default command for searching available man-pages. This would complement commands like `:Helptags` and fits well into commands with external dependencies like `:Locate`, `:Rg` and `:Commits`.

kattrali:
I went looking for something similar, and turned up this issue. For future travelers, this worked well for my purposes:

```vim
command! -bang -nargs=? Apropos
    \ call fzf#vim#grep('man --apropos '.fzf#shellescape(<q-args>), {
    \   'sink': {line -> execute('Man '. join(split(line, ' ')[:1],''))},
    \   'options': ['--preview', 'echo {1} | awk "{print \$1\$2}" | xargs man']
    \ }, <bang>0)
```

which is roughly equivalent to the command:

```sh
man -k $WORD | fzf --preview='man {1}{2}' | awk '{print $1;print $2}' | xargs man
```

Issue #818, opened 2019-06-01
Helptags does not respect default actions


adaszko:
- Category
    - [ ] Question
    - [X] Bug
    - [ ] Suggestion
- OS
    - [X] Linux
    - [X] macOS
    - [X] Windows
    - [X] Etc.
- Vim
    - [X] Vim
    - [X] Neovim

Thanks for a lovely Vim plugin!

The issue is that keys like CTRL-V, CTRL-T, CTRL-S do are not usable with `:Helptags`.  As far as I can see in the implementation, they're not handled https://github.com/junegunn/fzf.vim/blob/5c87ab86c0e7b0ec8a5edd39a4ce5500bb9e9333/autoload/fzf/vim.vim#L991-L1008

junegunn:
`Helptags` command simply executes `:help foo` to open help window. The split keys are not supported.

AlxHnr:
Why can't those be added? That would make it consistent with other commands like `:Tags`.

`:help ...` itself can be prefixed by `:vertical` and `:horizontal`.

junegunn:
Pull requests welcome.

PR #1520, opened 2024-01-13
Add an option to open commands in different ways (`:Helptags`, `:Manpages`)

jlbas:
This PR adds an option for commands that split a window to be opened in different ways. Currently this would only apply to `:Helptags` (#818).

I've also proposed a new `:Manpages` command (#1389) which would also make use of this option.

If this is something you'd like to incorporate I'd be happy to make a PR describing `g:fzf_cmd_mod` in [README-VIM](https://github.com/junegunn/fzf/blob/master/README-VIM.md#configuration).


meinzer1899:
Would be great to have that.

PR #1525, opened 2024-02-25
Improve :BCommits Preview

felixendres:
Reduce the preview of :BCommits to the relevant part, when used with visual selection. Only show the part of the diff that belongs to the selection.

That makes it very easy to quickly follow the evolution of a code block. I've used this feature over two month now, and it proved to be super useful.


junegunn:
Thanks!

PR #1529, opened 2024-03-01
[tags] Adds HTags command

### LINKED ISSUE #1524: `:Tags` is very slow on large tags file
- [x] I have fzf 0.30.0 or above

Hello,

Thanks for your beautiful tools :slightly_smiling_face: 

I have some repositories with larges tags file (130MB, few gigabytes) and when I use the `Tags` command to jump to the definition of a function, it takes a few seconds to process (up to 15 seconds on the [linux sources](https://github.com/torvalds/linux/archive/refs/heads/master.zip) ).

I had a quick look at the `fzf#vim#tags` function, and it looks like calling the `tags.pl` script takes quite some time (on large tags file).
I replaced the `tags.pl` with the following:
```sed
#!/usr/bin/sed -f
/^!/d
s/$/tags/
```
(tested only with a single tags file located at the root of my sources folder from where I open vim).
It is better, but still takes some time (compared to native `tjump`).
Then I tried to modify `fzf#vim#tags` to use `cat` and `s:tags_sink` to make up for it, it still takes 7-8 seconds to load something, while `tjump` is instantaneous (tags file size is 1.2GB)

I have reproduced the behavior with minimal vim configuration.

Do you have any idea on how to improve response time of tags search ?

nvincent-vossloh:
HTags use readtags to prefilter requested tags, this is useful for large tags file where the perl processing takes some time and the search is long.

fix #1524

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -100,6 +100,7 @@ let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:bin_dir = expand('<sfile>:p:h:h:h').'/bin/'
 let s:bin = {
 \ 'preview': s:bin_dir.'preview.sh',
+\ 'tagsprefilter': s:bin_dir.'tagsprefilter.sh',
 \ 'tags':    s:bin_dir.'tags.pl' }
 let s:TYPE = {'bool': type(0), 'dict': type({}), 'funcref': type(function('call')), 'string': type(''), 'list': type([])}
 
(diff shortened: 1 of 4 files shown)


junegunn:
I figured we didn't need a separate command for this and updated the Tags command to use readtags when a query string is provided. Let me know what you think. Thanks.

nvincent-vossloh:
Yes this is perfect. Glad you found a solution which did not require a separate command, I was not very satisfied with my solution.

PR #1546, opened 2024-05-25
fix environment variable reference in #3809

ykhan21:
This fix would allow nvim in cmd / powershell to use scoop_git_bash instead of wsl bash.
It should have been included in #3809.


junegunn:
Thanks!

PR #1547, opened 2024-06-04
prefer &shell to git bash on windows even if available

Konfekt:
calling git bash via cmd on windows is slow and opens a popup window

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -163,7 +163,7 @@ function! fzf#vim#with_preview(...)
     call remove(args, 0)
   endif
 
-  if !executable(s:bash())
+  if (has('win32') && &shell !~? '\v<(ba)?sh>') || !executable(s:bash())
     return spec
   endif
 


autoload/fzf/vim.vim
```diff
@@ -163,7 +163,7 @@ function! fzf#vim#with_preview(...)
     call remove(args, 0)
   endif
 
-  if !executable(s:bash())
+  if (has('win32') && &shell !~? '\v<(ba)?sh>') || !executable(s:bash())
```
junegunn, line 166: What exactly does this supposed to do?

The original condition checks if a `bash` is available, and if not, does not add preview options.

So you don't want preview window to appear? Is the title of the PR correct?

junegunn: Can you show me some screenshots before/after the patch?

Konfekt:
A lot of recent improvements on FZF in Git Bash likely resolved this ever more inscrutable issue

PR #1553, opened 2024-07-16
Jumps support preview window

zeminzhou:
Like this:
image

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1337,37 +1337,52 @@ endfunction
 " Jumps
 " ------------------------------------------------------------------
 function! s:jump_format(line)
-  return substitute(a:line, '[0-9]\+', '\=s:yellow(submatch(0), "Number")', '')
+  let line = substitute(a:line, '[0-9]\+', '\=s:yellow(submatch(0), "Number")', '')
+  let line = substitute(line, '\s.\{-}\ze:[0-9]\+:', '\=s:green(submatch(0), "Directory")', '')
+  let line = substitute(line, '\%(:[0-9]\+\)\+:', '\=s:black(submatch(0), "NonText")', '')
+  return line
 endfunction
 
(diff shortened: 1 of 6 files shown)


autoload/fzf/vim.vim
```diff
@@ -1367,7 +1367,7 @@ function! fzf#vim#jumps(...)
   return s:fzf('jumps', {
   \ 'source'  : extend(s:jumplist[0:0], map(s:jumplist[1:], 's:jump_format(v:val)')),
   \ 'sink*'   : s:function('s:jump_sink'),
-  \ 'options' : '+m -x --ansi --tiebreak=index --cycle --scroll-off 999 --sync --bind start:pos:'.current.' --tac --header-lines 1 --tiebreak=begin --prompt "Jumps> "',
+  \ 'options' : ['+m', '-x', '--ansi', '--tiebreak=index', '--cycle', '--scroll-off', '999', '--sync', '--bind', 'start:pos:'.current, '--tac', '--header-lines', '1', '--tiebreak=begin', '--prompt', 'Jumps> ', '--preview-window', '+{2}-/2'],
```
junegunn, line 1370: Some options are in `--foo=bar` format and some others are in `--foo bar`. We could make the code a bit more readable by consistently following the first style when it's viable.

```suggestion
  \ 'options' : ['+m', '-x', '--ansi', '--tiebreak=index', '--cycle', '--scroll-off=999', '--sync', '--bind=start:pos:'.current, '--tac', '--header-lines=1', '--tiebreak=begin', '--prompt', 'Jumps> ', '--preview-window', '+{2}-/2'],
```

junegunn:
Thanks, it kind of works. But when a jump position is in the current buffer, the last column is not the file path, but the line. It makes the preview command fail with a strange error message.

image

Let me see if there's a way to fix this problem.

junegunn:
Merged with some major overhaul. Thanks!

PR #1567, opened 2024-09-15
Update tag stack after BTags and Tags commands

### LINKED ISSUE #240: Save a tag jump into tag stack
- Category
    - [ ] Question
    - [ ] Bug
    - [x] Suggestion
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Vim
    - [x] Vim
    - [x] Neovim

It would be nice if :Tags saved a tag jump into stack so that I can jump back to original position just like the builtin C-] and C-t

pavoljuhas:
Let CTRL-T function just as after builtin tag jumps.

Close #240


junegunn:
Merged as dfeb973a544c878d6953285c2af2ff964f97edf5. Thanks!

PR #1582, opened 2025-02-21
autoload/fzf: get_git_root, wrap system call with fzf#with_sh

sim590:
This aims too fix an issue expressed over here:

https://github.com/junegunn/fzf/issues/3188

Instead of going with the fix attempt over here:

https://github.com/junegunn/fzf/pull/4264

From what I can tell, only git commands are broken in the setting where `&shell=pwsh` and it's because of the escaping of characters in the path. In pwsh, you only need to quote a string and the path should be resolved correctly.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -712,12 +712,9 @@ endfunction
 
 function! s:get_git_root(dir)
   let dir = len(a:dir) ? a:dir : substitute(split(expand('%:p:h'), '[/\\]\.git\([/\\]\|$\)')[0], '^fugitive://', '', '')
-  if &shell ==# 'cmd'
-    let dir = fzf#shellescape(dir)
-  elseif &shell ==# 'pwsh'
-    let dir = '"'.dir.'"'
-  endif
-  let root = systemlist('git -C ' . dir . ' rev-parse --show-toplevel')[0]
+  let [shell, shellslash, shellcmdflag, shellxquote] = fzf#use_sh()
+  let root = systemlist('git -C ' . fzf#shellescape(dir) . ' rev-parse --show-toplevel')[0]
+  let [&shell, &shellslash, &shellcmdflag, &shellxquote] = [shell, shellslash, shellcmdflag, shellxquote]
   return v:shell_error ? '' : (len(a:dir) ? fnamemodify(a:dir, ':p') : root)
 endfunction
 
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -712,7 +712,7 @@ endfunction
 
 function! s:get_git_root(dir)
   let dir = len(a:dir) ? a:dir : substitute(split(expand('%:p:h'), '[/\\]\.git\([/\\]\|$\)')[0], '^fugitive://', '', '')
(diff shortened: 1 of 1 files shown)


sim590:
So, I think that the culprit of the whole thing is that we are actually not using `cmd.exe` when calling `get_git_root`.

[In fzf repo](https://github.com/junegunn/fzf/blob/a24d274a3cdc680dd2d4e5664121e9727239f3b6/plugin/fzf.vim#L455), this is done by using `s:use_sh()`. But, this function is not exposed to this repo.

If we actually want to expose `s:use_sh` as an autoload public function, then we could just use `s:use_sh` before calling `systemlist` inside `get_git_root` and this would fix the issue also.

Whichever you prefer between my current PR and this other approach is good for me. I would actually prefer the latter, i.e. using `fzf#use_sh` if it was exposed like that. I think that it would be cleaner.

sim590:
I merged https://github.com/sim590/fzf.vim/commit/72a9d18ac84c168ffc6e08bea93c097324bc8658 in my branch. I think it's the better way.

The only thing, again, is how the two changes are going to be synced for the user considering that upgrading to the newer version of `fzf.vim` is going to break all Git commands using `get_git_root` if `fzf` repo is not also updated. So, how do we approach this?

sim590:
I'm really starting to spam this fzf and fzf.vim repos... Sorry.

I introduced the new more convenient function `fzf#with_sh` instead of exposing `s:use_sh`. So, the code now uses `fzf#with_sh`.

autoload/fzf/vim.vim
```diff
@@ -712,7 +712,7 @@ endfunction
 
 function! s:get_git_root(dir)
   let dir = len(a:dir) ? a:dir : substitute(split(expand('%:p:h'), '[/\\]\.git\([/\\]\|$\)')[0], '^fugitive://', '', '')
-  let root = systemlist('git -C ' . fzf#shellescape(dir) . ' rev-parse --show-toplevel')[0]
+  let root = fzf#with_sh({ -> systemlist('git -C ' . fzf#shellescape(dir) . ' rev-parse --show-toplevel')[0] })
```
junegunn, line 715: Would it help to just not use `fzf#shellescape` here?

```suggestion
  let root = systemlist('git -C ' . shellescape(dir) . ' rev-parse --show-toplevel')[0]
```

sim590: You are right. It fixes the issue. Do you want me to push this change and close the other PR?

junegunn: Yes, this seems to be the simplest and more to the point.

I can understand the confusion around `fzf#shellescape`. It was inteded to be used when building a command that is going to be run by fzf.

sim590: OK. Thanks for the clarification. I indeed got confused about this.

sim590:
I pushed [3a2926e](https://github.com/junegunn/fzf.vim/pull/1582/commits/3a2926e7c135c8fc5a5fe1f28a33f272f948cddc).

junegunn:
Merged, thank you very much for looking into the problem and fixing it.

sim590:
Of course! Thanks for merging!

PR #1588, opened 2025-04-14
Add workaround for `normal!` bug in nvim terminals

Mo-tmd:
fzf uses `normal! m'` for some reason. There's a known issue regarding `normal` usage in neovim. See
https://github.com/neovim/neovim/issues/4895 for more details

As a temporary (yeah, right) workaround, use setpos to set the bookmark instead.

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -369,7 +369,7 @@ function! s:action_for(key, ...)
   " errors. e.g. E471: Argument required: tab drop
   if !a:0
     if !edit
-      normal! m'
+      call setpos("''", getpos('.'))
       silent! call s:execute_silent(cmd)
     endif
   else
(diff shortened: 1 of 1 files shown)


junegunn:
I haven't used Neovim for a while. Can you explain what problem this solves? I'd like to know how to reproduce the problem you're experiencing.

Mo-tmd:
I get this error when I press Ctrl-(x/t/v) after starting `:Buffers` from a terminal window:
> Error detected while processing function 3[30]..<SNR>8_callback[25]..function 3[30]..<SNR>8_callback:
line   23:
Vim(normal):Can't re-enter normal mode from terminal mode

To reproduce:

- Start nvim
- `:e ~/.bashrc` # Or any file
- `:terminal`
- `:Buffers`
- `Ctrl-x`

I'm using nvim v0.10.1, although I doubt that is relevant. I reproduced using minimal `init.vim` with only package manager and fzf plugins.

junegunn:
Thanks, I can reproduce the problem and confirm this helps. Thanks for the patch!

Mo-tmd:
Thanks for merging so quickly :)

PR #1590, opened 2025-05-15
feat(preview): teach preview.sh to print directories

vegerot:


bin/preview.sh
```diff
@@ -66,20 +66,23 @@ if [[ -z "$BATCAT" ]]; then
...
 MIME=$(file --dereference --mime -- "$FILE")
-if [[ "${MIME:FILE_LENGTH}" =~ binary ]]; then
+if [[ "${MIME:FILE_LENGTH}" =~ binary ]] && [[ ! -d "$FILE" ]]; then
   echo "$MIME"
   exit 0
 fi
 
 DEFAULT_COMMAND="highlight -O ansi -l {} || coderay {} || rougify {} || cat {}"
+if [[ -d "$FILE" ]]; then
+  DEFAULT_COMMAND="tree -C -L2 {} || ls -l --color=always {}"
```
junegunn, line 84: Are we doing this in case `tree` is not available on the system? If that's the case, we might want to suppress the error message when it's non-existent.

junegunn: Nevermind, it's already done.

junegunn, line 84: How about just removing `-L2`?

vegerot: It's a tradeoff.  In the scrot, imagine if `Code` had many files in it.  Then the preview would only show one directory. The current implementation gives an overview of the files in the directory.

I'll leave it up to you as to which you prefer.

junegunn: The choice of depth 2 feels a bit arbitrary, but I'm fine with either.

junegunn:
Merged, thanks!

bin/preview.sh
vegerot: @junegunn Thanks for landing!  I picked 2 precisely because it looked good in the directory I was testing with :D

Realistically it'd make sense to have it be dynamic.  Let's say the selected directory has 100 children, then it should be -L1.  On the other hand, let's say the selected directory only has one file and it's a directory (I'm thinking of the common Java project where you have `src/com/my/company/`), then it should be `-L4`. 

The goal would have it be dynamic to fit the size of the preview window.  _First_, you optimize for showing all the children, and _second_ you optimize for expanding as many subdirectories as possible until the preview window is full.  For now, I just picked `-L2` because it looked good in the directory I was testing with lol

PR #1591, opened 2025-05-21
fix(preview): never preview binary files

vegerot:
I think this is a bug fix instead of a new feature.  The reason for this is
that the current behavior _overrides user choice_ of `FZF_PREVIEW_COMMAND`. No
matter what preview command they provide, we never run it on binary files.
However, if the user supplies _no_ preview command, then we'll run it through
`bat` which produces an error.  This sounds like a mistake.

This commit makes the behavior when no preview command is supplied the same as
the behavior when a preview command is given.

Before:
image

After:
image

### DIFF
--- a/bin/preview.sh
+++ b/bin/preview.sh
@@ -66,17 +66,17 @@ if [[ -z "$BATCAT" ]]; then
   fi
 fi
 
-if [ -z "$FZF_PREVIEW_COMMAND" ] && [ "${BATCAT:+x}" ] && [[ ! -d "$FILE" ]] ; then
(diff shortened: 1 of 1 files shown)


junegunn:
Do you prefer the "After:" output with mime types?

I didn't pay much attention to this because I was fine with the error message from bat, or maybe I even preferred it because it clearly showed the `[bat warning]` with color highlighting.

vegerot:
> I didn't pay much attention to this because I was fine with the error message from bat,

Good point.  Either way the current behavior is inconsistent.  I agree with the suggestion in your second message.

Issue #1158, opened 2020-11-01
[FR] add BMarks - marks in current buffer


laur89:
- [x] I have fzf 0.23.0 or above
- [x] I have read through https://github.com/junegunn/fzf.vim/blob/master/README.md
- [x] I have read through https://github.com/junegunn/fzf/blob/master/README-VIM.md
- [x] I have read through the manual page of fzf (`man fzf`)
- [x] I have searched through the existing issues

Feature request for `:BMarks`, similar to `:BTags` - to show marks in current buffer only.
Looks like `:Marks` shows marks across... all marks globally tracked?

im-n1:
For someone who uses marks regulary is `:Marks` unusable when you target current file. This is highly anticipated feature.

PR #1593, opened 2025-06-06
feat: Allow passing arguments of :marks to :Marks command

scottdlai:
This PR allows passing initial mark filtering of `:marks` command to `:Marks`. As a result, this also introduces new command `:BMarks` that only list marks in the buffer (i.e. marks `a-z`).

Screenshot 2025-06-05 at 19 12 48

Related issue: https://github.com/junegunn/fzf.vim/issues/1158

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -1362,11 +1362,22 @@ function! s:mark_sink(lines)
   execute 'normal! `'.matchstr(a:lines[1], '\S').'zz'
 endfunction
 
-function! fzf#vim#marks(...)
+function! fzf#vim#marks(initial_marks, ...)
   redir => cout
-  silent marks
+  if empty(a:initial_marks)
+    silent marks
+  else
+    execute 'silent! marks' a:initial_marks
+  endif
   redir END
+
   let list = split(cout, "\n")
+
+  " If 1st entry is not header, i.e. errors then just display header
+  if list[0] != 'mark line  col file/text'
+    let list = ['mark line  col file/text']
+  endif
+
   return s:fzf('marks', {
   \ 'source':  extend(list[0:0], map(list[1:], 's:format_mark(v:val)')),
   \ 'sink*':   s:function('s:mark_sink'),
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -68,6 +68,7 @@ call s:defs([
 \'command! -bar -bang Commands                                  call fzf#vim#commands(<bang>0)',
 \'command! -bar -bang Jumps                                     call fzf#vim#jumps(fzf#vim#with_preview({ "placeholder": "{2..4}"}), <bang>0)',
 \'command! -bar -bang -nargs=* Marks                            call fzf#vim#marks(<q-args>, <bang>0)',
+\'command! -bar -bang -nargs=* BMarks                           call fzf#vim#marks("abcdefghijklmnopqrstuvwxyz", <bang>0)',
 \'command! -bar -bang Changes                                   call fzf#vim#changes(<bang>0)',
 \'command! -bar -bang Helptags                                  call fzf#vim#helptags(fzf#vim#with_preview({ "placeholder": "--tag {2}:{3}:{4}" }), <bang>0)',
 \'command! -bar -bang Windows                                   call fzf#vim#windows(fzf#vim#with_preview({ "placeholder": "{2}" }), <bang>0)',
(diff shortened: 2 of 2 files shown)


autoload/fzf/vim.vim
```diff
@@ -1362,11 +1362,22 @@ function! s:mark_sink(lines)
   execute 'normal! `'.matchstr(a:lines[1], '\S').'zz'
 endfunction
 
-function! fzf#vim#marks(...)
+function! fzf#vim#marks(initial_marks, ...)
```
junegunn, line 1365: How about we make the API backward-compatible? That would require keeping the function signature instact and examining what's inside `...`.

autoload/fzf/vim.vim
```diff
@@ -1362,26 +1362,38 @@ function! s:mark_sink(lines)
...
+  let extra = []
+
+  if len(a:000) > 0
+    if type(a:000[0]) == type('')
+      let initial_marks = a:000[0]
+      let extra = a:000[1:]
+    elseif type(a:000[0]) == type({})
+      let extra = a:000
+    endif
+  endif
```
junegunn, line 1376: `len(a:000)` can be replaced with `a:0`, and `a:000[0]` can be replaced with `a:1`.

For example,

https://github.com/junegunn/fzf.vim/blob/3ed6ac56d03fff0389cd5b30dfd1e1f940fe91d0/autoload/fzf/vim.vim#L532-L533

scottdlai: thank you! I've updated the PR

autoload/fzf/vim.vim
```diff
@@ -1362,15 +1362,29 @@ function! s:mark_sink(lines)
...
+    execute 'silent! marks' initial_marks
+  endif
   redir END
+
   let list = split(cout, "\n")
+
+  " If first line is not the expected header, fallback to header only
+  if len(list) == 0 || list[0] !=# 'mark line  col file/text'
+    let list = ['mark line  col file/text']
+  endif
```
junegunn, line 1382: This happens on `BMarks` when the current file has no marks, right? If that's the case, why don't we just print a warning message and return here?

Some examples:

https://github.com/junegunn/fzf.vim/blob/3ed6ac56d03fff0389cd5b30dfd1e1f940fe91d0/autoload/fzf/vim.vim#L1212-L1214

https://github.com/junegunn/fzf.vim/blob/3ed6ac56d03fff0389cd5b30dfd1e1f940fe91d0/autoload/fzf/vim.vim#L1401-L1405

https://github.com/junegunn/fzf.vim/blob/3ed6ac56d03fff0389cd5b30dfd1e1f940fe91d0/autoload/fzf/vim.vim#L743-L748

scottdlai: sure I can make the behaviour similar to that. I was thinking of making the behaviour similar to `Rg` and `BLines` if nothing is found, then we show empty list

autoload/fzf/vim.vim
```diff
@@ -1362,15 +1362,29 @@ function! s:mark_sink(lines)
...
+    silent marks
+  else
+    execute 'silent! marks' initial_marks
+  endif
   redir END
+
   let list = split(cout, "\n")
+
+  " If first line is not the expected header, no marks found
+  if len(list) == 0 || list[0] !=# 'mark line  col file/text'
```
junegunn, line 1380: Wouldn't `empty(list)` be enough?

autoload/fzf/vim.vim
```diff
@@ -1362,15 +1362,29 @@ function! s:mark_sink(lines)
...
+    return s:warn('No marks found')
+  endif
+
   return s:fzf('marks', {
-  \ 'source':  extend(list[0:0], map(list[1:], 's:format_mark(v:val)')),
+  \ 'source':  extend([list[0]], map(list[1:], 's:format_mark(v:val)')),
```
junegunn, line 1385: Is this change necessary?

autoload/fzf/vim.vim
junegunn: Yeah, please do so. It would make the code simpler.

`Rg` is a bit different in that we don't know if the list is empty until we start fzf.

autoload/fzf/vim.vim
scottdlai: not necessarily, I just find it easier to to understand :D. I can revert that back

autoload/fzf/vim.vim
scottdlai: when no marks are found, `list` also contains error messages.

junegunn: I see. Since error messages start with `E`, we can simplify the code to:

```suggestion
  if empty(list) || list[0] =~# '^E'
```

autoload/fzf/vim.vim
junegunn: Yeah, I'd prefer to keep the diff minimal, and have unrelated cleanups in separate commits.

junegunn:
Merged, thanks!

PR #1594, opened 2025-06-18
Fix cmd hisotry correctly

phanen:
Feedkeys `nt` seems more correct than use `histadd`.

Related: https://github.com/ibhagwan/fzf-lua/pull/2117.


junegunn:
Thanks, could you be more specific about the issue this addresses?

junegunn:
Thanks, I could reproduce the problem and confirmed your patch fixes it.

PR #1604, opened 2025-12-06
avoid cooked terminal mode

Konfekt:
as `:help system()` states

		When prepended by |:silent| the terminal will not be set to
		cooked mode.  This is meant to be used for commands that do
		not need the user to type.  It avoids stray characters showing
		up on the screen which require |CTRL-L| to remove. >
			:silent let f = system('ls *.vim')

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -718,7 +718,7 @@ function! s:get_git_root(dir)
     return FugitiveWorkTree()
   endif
   let dir = len(a:dir) ? a:dir : substitute(split(expand('%:p:h'), '[/\\]\.git\([/\\]\|$\)')[0], '^fugitive://', '', '')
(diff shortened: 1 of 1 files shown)


junegunn:
Hi, could you explain what the motivation behind this is?

Konfekt:
As getting the git root dir doesn't need the user to type, keeping it silent seems to be the safer variant and just applying best Vim practice here

Konfekt:
I didn't check the code base for other occurrences of `system(list)()` yet as I'm awaiting feed back on the applicability of this Vim help recommendation

Konfekt:
I'm also wondering to what extent the help text is self-explanatory and applies to, say, `systemlist()` instead of `system()`. If not so much, one could pass that question to the Vim maintainers themselves first

junegunn:
I was just wondering if you actually had an issue because of the lack of `silent`.

junegunn:
Merged, thanks!

PR #1606, opened 2025-12-08
add  commit count to :GFiles to restrict to last <count> commits

Konfekt:
Let `:GFiles` take a count to restrict to those files altered in last <count> commits.
Count 0 corresponds to `:GFiles?` kept for backwards compatibility.
Builds in [this snippet](https://gist.github.com/Konfekt/429eb0f4d9a17e31235f49c4716ffddb)

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -713,6 +713,59 @@ endfunction
 " GFiles[?]
 " ------------------------------------------------------------------
 
+function! fzf#vim#gfiles(bang, count, args) abort
+  " When count is negative, fall back to the original GFiles behavior
+  if a:count < 0
+    return fzf#vim#gitfiles(a:args, fzf#vim#with_preview(a:args == '?' ? { 'placeholder': '' } : {}), a:bang)
+  endif
+
+  let root = s:get_git_root('')
+  if empty(root)
+    return s:warn('Not in git repo')
+  endif
+
+  " Count 0 behaves like :GFiles? (git status)
+  if a:count == 0
+    return fzf#vim#gitfiles('?', fzf#vim#with_preview({ 'placeholder': '' }), a:bang)
+  endif
+
+  " Positive count: files changed in the last {count} commits
+  let filter = s:is_win ? 'findstr /v "^$"' : 'sed -e /^$/d'
+  let prefix = 'git -C ' . fzf#shellescape(root) . ' '
+  let source = prefix . 'log HEAD --max-count='.a:count.
+        \ ' --diff-filter=MA --name-only --pretty=format: | '.filter
+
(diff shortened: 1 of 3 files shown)


Konfekt:
Glad to see you generalizing this in https://github.com/junegunn/fzf-git.sh/pull/104 

> "last N" commits, which to me feels a bit arbitrary and restrictive.

I agree that there's only so much one can restrict to given a single number (though often the more recent commits are usually the ones one is most concerned with); so it came to my mind that Vim allows for more general contiguous ranges. Would a range`n,m` for the files changed in between the `n`th and `m`th commit make this sufficiently generally applicable? That's also likely as general as it gets to be able to display the diffs of these files changed in a set of commits.

junegunn:
Thanks.

> Would a rangen,m for the files changed in between the nth and mth commit make this sufficiently generally applicable?

I'm not sure. The commits can be non-contiguous? It feels like we might be trying to overload the `:GFiles` command a bit too much, which could make its behavior harder to reason about.

Defining a command that does what the suggested `:[X]Gfiles` isn't particularly complicated. So until we have a clearer idea of what we want to do, users can use a custom command tailored to their needs. For example:

```vim
command! -nargs=1 GDT call fzf#run(fzf#wrap({
  \ 'source': printf("git diff-tree --no-commit-id --name-only -r 'HEAD~%d..HEAD'", <q-args>),
  \ 'options': ['--multi', '--preview', printf('git diff --color=always "HEAD~%d..HEAD" -- {}', <q-args>)]}))
```

Konfekt:
> commits can be non-contiguous?

Maybe, but at least with a `HEAD~n...HEAD~m` there's a chance; necessary, not sufficient a condition as in math speak.

> It feels like we might be trying to overload the :GFiles command a bit too much

I had that feeling for `:Gfiles?` which demands special parsing and seems more coherently labeled `0Gfiles`.

>  users can use a custom command tailored to their needs

Wow, that's indeed much simpler than [my involved snippet](https://gist.github.com/Konfekt/429eb0f4d9a17e31235f49c4716ffddb) (@jparise might be interested as well). 

I will try to package it into a command taking the `n,m` range and falling back to `:Gfiles` for no range given; maybe that's where much code went into

junegunn:
> Maybe, but at least with a HEAD~n...HEAD~m there's a chance

Oh, I mean, there can be cases where the user might want to see the list of modified files from non-continguous commits, which is supported in https://github.com/junegunn/fzf-git.sh/pull/104. e.g. `<ctrl-g><ctrl-h> ... select commits with <tab>s ... press <alt-f> to list files`.

> I had that feeling for `:Gfiles?` which demands special parsing and seems more coherently labeled `0Gfiles`.

Right. Probably not the best decision I made :)

PR #1611, opened 2026-04-01
Fix E728 in fzf#vim#colors() when called with a spec dict

### LINKED ISSUE #1610: E728 in fzf#vim#colors() when first arg is a dictionary
## Bug

`fzf#vim#colors()` throws `E728: Using a Dictionary as a Number` on line 614 of `autoload/fzf/vim.vim` when called with a spec dict as the first argument.

## Cause

```vim
if !a:1 " We can't set up IPC in fullscreen mode in Vim
```

This assumes `a:1` is always a number (`<bang>0` from the `:Colors` command), but the function's public API accepts `([spec dict], [fullscreen bool])` — so `a:1` can be a dictionary.

## Steps to reproduce

```vim
:call fzf#vim#colors({})
```

## Suggested fix

```vim
if a:0 && type(a:1) != s:TYPE.dict && !a:1
```

This matches how `s:fzf()` (line 322) already handles the same varargs pattern.

## Version

Latest master (`34a564c81f`)

Introduced in `04bfa81157` ([Colors] Change color scheme on the fly (experimental)).

MenkeTechnologies:
## Problem

`fzf#vim#colors()` throws `E728: Using a Dictionary as a Number` when called with a spec dict as the first argument:

```vim
command! -bang Colors call fzf#vim#colors({'left': '20', 'options': '--reverse' }, <bang>0)
```

Line 614 does `if !a:1` assuming `a:1` is always a number, but the documented API is `fzf#vim#colors([spec dict], [fullscreen bool])`.

## Fix

Add a type check before the boolean test, matching how `s:fzf()` already handles the same varargs pattern:

```vim
if a:0 && type(a:1) != s:TYPE.dict && !a:1
```

Fixes #1610

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -611,7 +611,7 @@ function! fzf#vim#colors(...)
   \ 'options': ['+m', '--prompt', 'Colors> ']
   \}
 
-  if !a:1 " We can't set up IPC in fullscreen mode in Vim
+  if a:0 && type(a:1) != s:TYPE.dict && !a:1 " We can't set up IPC in fullscreen mode in Vim
     let fifo = fzf#vim#ipc#start({ msg -> execute('colo '.msg) })
     if len(fifo)
       call extend(spec.options, ['--no-tmux', '--no-padding', '--no-margin', '--bind', 'focus:execute-silent:echo {} > '.fifo])


autoload/fzf/vim.vim
```diff
@@ -611,7 +611,7 @@ function! fzf#vim#colors(...)
   \ 'options': ['+m', '--prompt', 'Colors> ']
   \}
 
-  if !a:1 " We can't set up IPC in fullscreen mode in Vim
+  if a:0 && type(a:1) != s:TYPE.dict && !a:1 " We can't set up IPC in fullscreen mode in Vim
```
junegunn, line 614: Since the function can take at most 2 arguments, the fullscreen flag may not be a:1 but a:2.

What do you think?

```suggestion
  let fullscreen = a:0 && type(a:000[-1]) != s:TYPE.dict && a:000[-1]
  if !fullscreen " We can't set up IPC in fullscreen mode in Vim
```

Test cases:

```vim
call fzf#vim#colors()
call fzf#vim#colors(0)
call fzf#vim#colors(1)
call fzf#vim#colors({})
call fzf#vim#colors({}, 0)
call fzf#vim#colors({}, 1)
```

MenkeTechnologies: looks resilient enough.

PR #1619, opened 2026-06-01
Show lua colorschemes in :Colors command

sach1t:
Problem:
The `:Colors` command only looks for color schemes written in vimscript and not any Neovim lua color schemes.

Fix:
If available use`getcompletion` to get the list of available color schemes. This should be more accurate than globbing the paths.  If `getcompletion` isn't available then fallback to the prior behavior with added support for lua colorschemes.

I would be in favor of removing the fallback completely and just using `getcompletion`, but that would drop support for Vim versions prior to 7.4 when `getcompletion` was added.

Testing:
- [x] neovim 0.12
- [x] vim 9.1

### DIFF
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -688,11 +688,7 @@ function! s:colors_exit(code)
 endfunction
 
 function! fzf#vim#colors(...)
-  let colors = split(globpath(&rtp, "colors/*.vim"), "\n")
-  if has('packages')
-    let colors += split(globpath(&packpath, "pack/*/opt/*/colors/*.vim"), "\n")
-  endif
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks!

> I would be in favor of removing the fallback completely and just using `getcompletion`, but that would drop support for Vim versions prior to 7.4 when `getcompletion` was added.

7.4 was released in 2013, let's drop it!

sach1t:
I removed the fallback.

Sorry I didn't realize but `getcompletion` was patch 7.4.2011, so it was actually rolled into [Vim 8](https://groups.google.com/g/vim_announce/c/EKTuhjF3ET0/m/cP1JQcZ4AwAJ) which is 10 years old (2016). It should still be safe to drop.

junegunn:
Merged, thanks!

